PR #4, opened 2015-03-03
Removes unneeded quote symbol

potherca:
The quote symbol `"` after 

> But they have to be initialised at the time of declaration.

does not seem to belong to an opening quote symbol.


rafaparadela:
I can't automatically merge it, since the previous PR merged changed the same line. I'm editing it manually. Thank you for your contribution @potherca

PR #5, opened 2015-03-04
Person case class definition

padilo:
Missing Person case class definition and default value definition to answer the exercises


rafaparadela:
@padilo Thank you for your contribution. ;-)

PR #9, opened 2015-03-04
adding assignment example for tuples

josephpconley:
I thought it might be useful to show an example where a tuple can be quickly assigned to multiple variables, i.e.

    val student = ("Sean Rogers", 21, 3.5)
    val (name, age, gpa) = student

### DIFF
--- a/app/json/tuples.json
+++ b/app/json/tuples.json
@@ -20,6 +20,16 @@
       "postparagraph": ""
     },
     {
+      "preparagraph": "You can assign multiple variables at once using tuples:",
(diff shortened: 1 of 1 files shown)


rafaparadela:
Yes!! Example pretty illustrative. Thank you @josephpconley . ;-)

josephpconley:
One thing to note, I had already finished the tuples section before this PR, now when I go back into it the sequencing of answers is messed up (i.e. my answers for the last question showed up in this new question).  Just an FYI

PR #15, opened 2015-03-05
Update maps.json

alien11689:


rafaparadela:
@alien11689 Thank you for your contribution. I think that the way you've addressed the problem of accessing to not found key, is valid. But doesn't comply with the philosophy that each exercise is an assertion that must be met. What do you think @raulraja ?

The original exercise:
```
val myMap = Map("OH" -> "Ohio", "WI" -> "Wisconsin", "IA" -> "Iowa")

intercept[NoSuchElementException] {
  myMap("MI") should be()
}
```

The new proposal:
```
val myMap = Map("OH" -> "Ohio", "WI" -> "Wisconsin", "IA" -> "Iowa")
val aNewMap = myMap - "OH"

intercept[NoSuchElementException] {
  aNewMap(__)
}
```

raulraja:
@rafaparadela @alien11689 I think this will prove this case better:

```scala
val myMap = Map("OH" -> "Ohio", "WI" -> "Wisconsin", "IA" -> "Iowa")
var blewWithException = true
intercept[NoSuchElementException] {
  myMap("MI")
  blewWithException = false
}
blewWithException should be(_)
```

`blewWithException` should remain `true` since `myMap("MI")` will raise the exception before `blewWithException` mutates.

rafaparadela:
@raulraja This solution looks good to me.
@alien11689 Could you address this change in order to preserve your contribution, please? Thank you.

alien11689:
Ok, I've fixed it.

rafaparadela:
Cool. Thank you @alien11689

PR #16, opened 2015-03-06
fixed closure sample

michaelmosmann:
A closure captures the values from surrounding context just once. It will not change anything if someone changes a value after the capture, so the samples are a bit misleading...

### DIFF
--- a/app/json/higherorderfunctions.json
+++ b/app/json/higherorderfunctions.json
@@ -23,20 +23,19 @@
       "postparagraph": ""
     },
     {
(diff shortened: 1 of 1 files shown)


rafaparadela:
AFAIK in each closure invocation, the external variable is evaluated.

```
scala> var incrementer = 1
incrementer: Int = 1

scala> def closure = { x: Int => x + incrementer }
closure: Int => Int

scala> val result1 = closure(10)
result1: Int = 11

scala> incrementer = 2
incrementer: Int = 2

scala> val result2 = closure(10)
result2: Int = 12
```

Does closure work differently at REPL?

michaelmosmann:
Oh.. I was wrong. Had a typo in my test code. So you should not merge this!

rafaparadela:
@michaelmosmann actually this issue has sparked a debate in our office. :-)

michaelmosmann:
If you try this with java8, the compiler will give you an error, because he makes "incrementer" final.. and an second assignment will fail. Maybe this is what i would expect in scala too.. :)

rafaparadela:
:-)

PR #19, opened 2015-03-06
Fix 3rd secondElement definition in patternmatching.json

PierreMage:
In:

    val secondElement = List(1) match {
      case x :: y :: xs => xs
      case _ => 0
    }

the second line should be:

      case x :: y :: xs => y


rafaparadela:
Makes sense. Thank you @PierreMage

PR #21, opened 2015-03-08
Fixes 'Movie' pattern matching case 1932 year

isabelabel:


rafaparadela:
Makes sense. Thank you @isabelabel

isabelabel:
You're welcome @rafaparadela

PR #22, opened 2015-03-10
case classes are serializable

padilo:
Case classes are Serializable.

btw, with koans generator is slightly easier :P


rafaparadela:
Sounds good. Thank you @padilo

PR #23, opened 2015-03-10
For translates to flatMap, map & filter

padilo:
An example on how for can be translated to flatMap, map and filter

### DIFF
--- a/app/json/forexpressions.json
+++ b/app/json/forexpressions.json
@@ -28,12 +28,9 @@
     },
     {
       "preparagraph": "Using 'for' we can make more readable code",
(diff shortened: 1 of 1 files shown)


rafaparadela:
Really interesting exercise @padilo, but I think that last part of the code could be removed since doesn't conform to the koans philosophy which always ask for right side of the assertion. Also, the first part is enough clarifier to understand for-comprehension.

Could you please leave just the part below?

```
val nums = List(List(1), List(2), List(3), List(4), List(5))

val result = for {
  numList <- nums
  num <- numList
  if(num%2 == 0)
} yield(num)

result should be (__)
```

Thank you very much for these great contributions, dude.

padilo:
I removed the last, but I still leave the second part resolved (and simplified) just to show how the same can be done using flatmap and filter.

If do you prefer to remove it anyway I will not insist on this :)

rafaparadela:
It's ok. Thanks. ;-)

Issue #18, opened 2015-03-06
sameElements compares the elements, not just the number of elements


davidallsopp:
http://scala-exercises.47deg.com/koans#iterables says that "sameElements will return true if the two iterables have the same number of elements"

That's a necessary, but not sufficient condition. The elements have to be the same, too. Whether the ordering matters depends on the collection:

scala> List(1,2,3) sameElements List(1,2,3)
res13: Boolean = true

scala> List(1,2,3) sameElements List(3,2,1)
res14: Boolean = false

scala> Set(1,2,3) sameElements Set(3,2,1)
res15: Boolean = false                                                            // WHAT?

scala> Set(1,2,3,4,5,6) sameElements Set(6,5,4,3,2,1)
res16: Boolean = true

Also, sameElements for Set produces unexpected results - the behaviour changes depending on the size of the Set, as you can see above! This is arguably a bug in the Scala library: https://groups.google.com/forum/#!topic/scala-debate/je4vZbR3WbA

rafaparadela:
Thank you @davidallsopp , I didn't know about this bug. You are reporting several issues. I would like that your contributions gain recognition. Why don't you submit a new "pull request" with your fix for this particular issue?

- That'd the JSON to edit: https://github.com/47deg/scala-exercises/edit/master/app/json/iterables.json

- And this is a very simple tool I made, to edit exercises: http://rafaparadela.github.io/koans-generator/ (You can paste on "JSON column", edit it, and preview it).

Thanks a lot, dude!

rafaparadela:
Fixed!

xaniasd:
The _sameElements_ part of [Iterables](https://www.scala-exercises.org/std_lib/iterables), in its current form is confusing. I see that the explanation submitted in this [PR](https://github.com/scala-exercises/scala-exercises/pull/24) is missing. Perhaps removed by accident?

PR #24, opened 2015-03-10
More precise explanation of sameElements, addressing issue #18

davidallsopp:
Added example showing that sameElements returns false when elements of the lists are in a different order.
Added example and explanation of the surprising behaviour of sameElements with small Sets.
Addresses issue #18

### DIFF
--- a/app/json/iterables.json
+++ b/app/json/iterables.json
@@ -128,13 +128,15 @@
       "postparagraph": ""
     },
     {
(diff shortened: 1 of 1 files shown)


rafaparadela:
Nice @davidallsopp !! This exercise is enough clarifier to understand how "sameElements" works.
Thank you a lot.

PR #28, opened 2015-03-20
Update traversables.json

blueant:
take function in Stream should return a Stream object instead of List object

### DIFF
--- a/app/json/traversables.json
+++ b/app/json/traversables.json
@@ -150,12 +150,11 @@
     },
     {
       "preparagraph": "`toStream` will convert any *Traversable* to a `Stream` which is a lazy list where elements are evaluated as they are needed.",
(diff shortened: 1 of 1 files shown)


rafaparadela:
@blueant Although a `Stream` object is returned by `take`, `Stream` is a lazy list indeed. In fact, in Scala, a `Stream` is a `List` whose tail is a lazy val. In this case when `ScalaTest` compares the assertion `should be`, the value of the tail is computed. In the same way as it would be evaluated with `List(4, 6, 7) == result.take(3)` or whatever other command that demands the tail value.
In the other hand, `Stream(4, ?)` is a representation of the lazy list, not comparable since symbol `?` is can't be resolved.
Thank you for your contribution, but in this case, we can't merge it.

PR #29, opened 2015-03-21
Update range.json

fada21:
Now first two examples force to think more and are more illustrative IMO.


rafaparadela:
That's amazing. Thank you for your contribution.

fada21:
I thank you. This tutorial is awesome. Best quick revision after Functional
classes at Coursera. Just keep it simple and essential as it is now.

--
Łukasz

rafaparadela:
:-)

PR #30, opened 2015-04-01
Maps - Immutability

DrMabulle:
Insists on the fact that the Map is immutable


rafaparadela:
Makes sense. Thank you for your contribution. ;-) Shipping the boat...

PR #32, opened 2015-05-19
Update manifest.json

kasured:
I have presented a reader with some deeper insight into the initial reasons standing behind the Manifest feature covered with some simple code example. Also, please, pay attention to the Funtion1[Int, Int] instead of Funtion1[int, int] as the output of name[Int => Int]. Thxz.


rafaparadela:
;-)

PR #33, opened 2015-05-20
Update extractors.json to provide more details.

kasured:
I reckon the actual code expansion by compiler that will occur on the pattern matching will put more light on how stuff works actually and when apply/unapply calls happen.


rafaparadela:
Makes sense. Thank you @kasured

PR #34, opened 2015-05-25
Update traits.json

baero:
minor edit


rafaparadela:
Thank you @baero !! Shipping!

PR #36, opened 2015-06-02
Update valandvar.json

BenjaminChe:
Typo correction in preparagraph 1

>"will result in al reassignment" 
>"will result in a reassignment"

Additional content elaborating on the ability to modify the internal state of a val added.
>"postparagraph": ""
>"postparagraph": "Let us consider an Array being declared as `val`.\n\n```\nval stringArray:Array[String] = new Array(6)\n```\n\n  The stringArray can be modified (the internal state of a val can be changed), but the reference/assignment cannot be modified to point to another Array.\n\n```\nstringArray = new Array(33)\n```\n\nwill result in a reassignment to val error, but\n\n```\nstringArray(3) = "foo"\n```\n\nwill not result in the error."
+    }

### DIFF
--- a/app/json/valandvar.json
+++ b/app/json/valandvar.json
@@ -2,7 +2,7 @@
   "title": "Val and Var",
   "modules": [
     {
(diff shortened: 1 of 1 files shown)


rafaparadela:
Thank you @BenjaminChe . 

The typo correction is ok. But the "postparagraph" clarification should be edited due to issue. Double quotes for "foo" should be scaped \"foo\", something like that:

```
"Let us consider an Array being declared as `val`.\n\n```\nval stringArray:Array[String] = new Array(6)\n```\nThe `stringArray` can be modified, but the reference cannot be modified to point to another Array.\n\n```\nstringArray = new Array(33)\n```\nwill result in a reassignment to val error, but\n```\nstringArray(3) = \"foo\"\n```\nwill not result in the error."
```

Anyway, you can use my koan generator for live previewing: http://rafaparadela.github.io/koans-generator/

BenjaminChe:
Thanks for the koan generator!   I applied your formatting fix and clarified some of the additional content.

rafaparadela:
Cool... merging!

PR #37, opened 2015-06-02
Added comment mySet.sum

d1egoaz:
Added the comment that you can also sum elements in a Set with 
mySet.sum


rafaparadela:
Makes sense. Thank you @d1egoaz

PR #38, opened 2015-06-03
.flatten instead of .flatMap(x => x) in For Expressions

d1egoaz:
In the last part, I think is clearer to use
``` scala
nums.flatten.filter(_%2==0)
```
rather than 

``` scala
nums.flatMap(numList=>numList).filter(_%2==0)
```

I didn't replace that piece of code but added the flatten part at the end


rafaparadela:
@d1egoaz Cool! Thank you.

PR #39, opened 2015-06-03
zipall E.g. is fixed added E.g. of contrary case

d1egoaz:
E.g says that
```scala
Iterable(x1, x2, x3) zipAll (Iterable(y1, y2), x, y) will return ((x1,y1), (x2, y2, y))
```
but must return
```scala
Iterable(x1, x2, x3) zipAll (Iterable(y1, y2), x, y) will return ((x1,y1), (x2, y2), (x3, y)))
```

Additionaly, this example is added, it shows the contrary case

```scala
va xs = List(3, 5)
val ys = List("Bob", "Ann", "Stella")

(xs zipAll(ys, -1, "?")) should be(List((3,"Bob"), (5,"Ann"), (-1,"Stella")))
```

### DIFF
--- a/app/json/iterables.json
+++ b/app/json/iterables.json
@@ -105,14 +105,19 @@
       "postparagraph": ""
     },
     {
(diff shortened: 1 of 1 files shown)


rafaparadela:
Makes sense. 

Thank you very much.

PR #40, opened 2015-06-04
Change Int type to underscore

stefaneng:
The description of the code block says that the type parameters can be blocked out, but the `isInstanceOf` was using `Function1[Int,Int]` instead of `Function1[_, _]`


rafaparadela:
Makes sense. Thank you for contributing!

PR #41, opened 2015-06-10
Add notes on manifest

vincentlaucy:
```
In Scala 2.10, scala.reflect.ClassManifests are deprecated, and it is planned to deprecate scala.reflect.Manifest in favor of TypeTags and ClassTags in an upcoming point release. Thus, it is advisable to migrate any Manifest-based APIs to use Tags.
```
from http://docs.scala-lang.org/overviews/reflection/typetags-manifests.html


rafaparadela:
Makes sense. Thank you.

PR #43, opened 2015-07-10
add collection style operations to options

natansil:
add collection style operations like map and fold to options


rafaparadela:
These koans are great. Thank you.

PR #45, opened 2015-07-26
Parent classes koan fix

fada21:
Your koans serve me as a scala reference. I want to see them working and updated. Error reminded me about catch22 so I changed example a bit, I hope you like it.


rafaparadela:
Maybe your approach is clearer. Thank you!

PR #50, opened 2015-08-17
Update higherorderfunctions.json

RomanVlasenko:
Code example is not corresponding to description


rafaparadela:
Makes sense! Thank you.

PR #51, opened 2015-08-21
Check is bit ambiguous. Fixed it by passing right argument

mathant:
It should accept any 'Long' value for hours. or we need to pass a fixed value which we can verify.


rafaparadela:
This issue has been resolved in a previous PR. Thank you very much.

PR #62, opened 2015-10-07
SKO-32 Integrate new layout

rafaparadela:
This PR integrates the new layout for home page.

@raulraja Could you take a look please? Thank you.


scala-exercises-server/app/assets/stylesheets/path/css-directory
```diff
@@ -0,0 +1,23 @@
+/*
+Errno::ENOENT: No such file or directory @ rb_sysopen - path/sass-directory
+
+Backtrace:
+/Users/israelperezglez/.rvm/gems/ruby-2.1.1/gems/sass-3.4.16/lib/sass/plugin/compiler.rb:482:in `read'
```
raulraja: I'm guessing all of these files pointing to @israelperezglez mahcine shouldn't be checked in and perhaps where not added to .gitignore?

rafaparadela:
@FPerezP @raulraja Addressed all the comments related to Bootstrap removal.
Is ready to be merged?

scala-exercises-server/public/javascripts/parallax.js
```diff
@@ -0,0 +1,358 @@
+/*!
+ * parallax.js v1.3.1 (http://pixelcog.github.io/parallax.js/)
```
raulraja: I believe Parallax is also available as a webjar

raulraja:
Let's try to see if we can include Parallax via webjars or CDN but not importing it's sources. Once that is addressed, this LGTM! Good work!

rafaparadela:
Yes, webjars has `"org.webjars.bower" % "parallax.js" % "1.3.2"`. But the interactivity with javascript (scala.js) in this particular screen, is still pending. For this reason I didn't refactorize this library. 
Once the client-side is implemented (with scala-js), my intention is to package every webjars dependency this way:

```
skip in packageJSDependencies := false
```

And import the generated js (minified) as:

```
<script type="text/javascript" src="./target/scala-2.11/scala-exercises-jsdeps.js"></script>
```

I explained this tecnique in the scala-js-demo project. http://47deg.github.io/scala-js-demo/ (Slide 6).

The `bootstrap` case is different since this particular library is needed in the server side to compile sass files.

Let me address the js-dependencies refactorization in the next PR, when I have to develop several features at home screen as animating of the progress bars, etc.

Makes sense?

raulraja:
sounds good!

Issue #61, opened 2015-10-07
Add continuous integration to the project


rafaparadela:

fedefernandez:
Added integration with Travis

PR #64, opened 2015-11-10
Add continuous integration

fedefernandez:
Issue #61 

Actually it only compiles the code because the tests need a database. We can add a new ticket in a future to add the tests execution in the CI

Please @rafaparadela could you review? Thanks!


rafaparadela:
LGTM

Issue #71, opened 2015-11-17
Set a initial approach to work on follow up exercises


rafaparadela:
Create:
- Followup Model
- Followup Store
- Evolution to create Followup table
- Followup Service
- Followup Controller
- Followup Routes

rafaparadela:
Won't fix

PR #73, opened 2015-11-20
#71 Create the logic to show the follow up

rafaparadela:
This PR adds:

- Model and Store for followup entities
- Services to add and retrieve followup data
- Controller and endpoints to handle followup info.
- Changes in home template to show the current followup of the logged user.
- New evolution to have sample data
- New sample section with two categories

An example: [http://cdn.rafp.es/akN.png](http://cdn.rafp.es/akN.png)

@raulraja could you review please? Thank you.


raulraja:
Closing since we are considering a better approach to account for progress and exercise tracking as POC in the incubating v2

rafaparadela:
Sounds good

PR #76, opened 2015-11-21
Reformat code in "Constructors"

idelpivnitskiy:
I've just added missing spaces before '{' and '+' symbols and removed some unnecessary empty lines.


rafaparadela:
Cool, thanks

PR #77, opened 2015-11-22
Update higherorderfunctions.json

Phil-Ba:
Moved part about currying to partial applied functions.


rafaparadela:
Makes sense. Thank you.

PR #78, opened 2015-11-22
Update partiallyappliedfunctions.json

Phil-Ba:
Moved currying part from higher order functions. Changed description text a bit. Example showcasing how you can apply a single value to a
curreid function.


rafaparadela:
Makes sense. Thank you for enriching the concept.

PR #79, opened 2015-11-26
To/until in range koans

simplyianm:
Show the to/until syntax in range koans.


rafaparadela:
They are both good exercises. Thank you.

PR #81, opened 2015-12-07
It may be interessting to precise the type of result1

brainztorm:
val number: Option[Int] = Some(3)
val result1 = number.map(_ * 1.5) // The type of result1 is now Option[Double]

### DIFF
--- a/app/json/options.json
+++ b/app/json/options.json
@@ -56,13 +56,13 @@
       "postparagraph": ""
     },
     {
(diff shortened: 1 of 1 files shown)


rafaparadela:
Due to it's a clarification related to the previous exercises, maybe it makes sense to add it as postparagraph of this particular one, isn't it?
Anyway thank you a lot for your contribution.

brainztorm:
You're right @rafaparadela, i made the modification.
Have a good day and thanks for those exercise !

rafaparadela:
Thank you @brainztorm

PR #82, opened 2015-12-16
Use local H2 DB when running application server

andyscott:
This updates the project to use a local H2 DB (persisted to `~/.scala-exercises/db`) when `sbt run` is performed. Previously, Postgres was required to be running locally.


rafaparadela:
LGTM!

PR #83, opened 2015-12-16
Add sbt-scalariform

### LINKED ISSUE #60: Add Scalariform to enforce style and indentation in Compile phase

andyscott:
This would close #60.

Needs to be rebased after changes from #82 are taken care of.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,12 +1,27 @@
 import play.PlayImport._
 import sbt.Project.projectToRef
+import scalariform.formatter.preferences._
 
 lazy val clients = Seq(scalaExercisesClient)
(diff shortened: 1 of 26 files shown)


andyscott:
@raulraja is this what you had in mind?

raulraja:
Yes though we may want to tweak the settings down the road to what we are used to. For the time being this is good. We are also not married to a particular style so I'd love to see reasonings around why certain styles may work better than others. LGTM and thanks for doing this! great stuff!

andyscott:
I'm not set on a particular style either-- my biggest concern is just having a consistent style. Additionally, scalariform only affords you certain customizations, so we may be limited by that.

PR #84, opened 2015-12-17
Build file cleanup

andyscott:
Some preliminary build file cleanup. More to come, but probably not until way down the road when it's worth the effort.


raulraja:
Cuttlefish LGTM! 
south-park-s15e01c11-cuttlefish-is-not-sitting-well-16x9

PR #85, opened 2015-12-23
Fixed a small grammatical error in the val var lesson.

benslant:
Just a small grammar fix. "Any attempt to *[do]* so will result in a reassignment"


rafaparadela:
Thank you @benslant

PR #90, opened 2015-12-29
Fixed issues with partiallyappliedfunctions.json

dcy2003:
Fixed issue in which multiplyCurriedFour(4) incorrectly validates against 12 instead of the correct answer of 16.
Also fixed issue with 'val multiplyCurriedFour=multiplyCurried(4) _' which fails in Scala REPL.  
Replaced with 'val multiplyCurriedFour=multiplyCurried(4)'.


rafaparadela:
You are right. Thank you @dcy2003

PR #91, opened 2015-12-30
Coloring code blocks

rafaparadela:
This PR adds Highlight.js to colour blocks of code.

@raulraja @andyscott Could you review please? Thank you.

Example


andyscott:
Looks good to me. 👍🏼

scala-exercises-client/src/main/scala/scripts/ExercisesJS.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.scalajs.dom
+import org.scalajs.jquery.{ jQuery ⇒ $ }
+
+object ExercisesJS extends js.JSApp {
+
+  def main(): Unit = {
+    runHighlight
+  }
+
+  def runHighlight = $("pre code").each({ (f: Any, code: dom.Element) ⇒
```
raulraja: If we we are not using `f : Any` we should probably wildcard it `_`

raulraja:
LGTM

PR #92, opened 2015-12-30
Update sequencesandarrays.json

dcy2003:
The REPL gives a List, so including an alternative accepted response for correctness.


rafaparadela:
Makes sense. Thank you!

PR #93, opened 2015-12-30
Update traversables.json

dcy2003:
Incorrect wording.  I suspect copy/paste error, but "first" should be "last" in this example.


andyscott:
Looks good to me. Thanks!

PR #95, opened 2016-01-02
Exercise evaluation 

rafaparadela:
This PR adds:
- Implementation of the logic to add "inputs fields" in the exercise code.
- The logic to calculate when should occur the call to evaluate the exercise.

In summary, it works as shown: http://rafp.es/1TuELka

@raulraja @andyscott Could you review please? Thank you.

### DIFF
--- a/scala-exercises-client/src/main/scala/config/Routes.scala
+++ b/scala-exercises-client/src/main/scala/config/Routes.scala
@@ -2,13 +2,9 @@ package config
 
 object Routes {
 
-  object Users {
-    val base = "/users"
-    def all = base + "/all"
-    def create = base + "/create"
-    def update(id: Long) = base + s"/update/$id"
-    def delete(id: Long) = base + s"/delete/$id"
-    def clear = base + "/clear"
+  object Exercises {
+    val base = "/exercises/sections"
+    def evaluate(section: String, category: String) = base + s"/$section/$category/evaluate"
   }
 
 }
--- a/scala-exercises-client/src/main/scala/config/Routes.scala
+++ b/scala-exercises-client/src/main/scala/config/Routes.scala
@@ -1,10 +1,19 @@
 package config
 
+import scala.scalajs.js
+
 object Routes {
+  private val controllers = js.Dynamic.global.jsRoutes.controllers
 
   object Exercises {
-    val base = "/exercises/sections"
-    def evaluate(section: String, category: String) = base + s"/$section/$category/evaluate"
+    def sections: String =
+      controllers.ExercisesController.sections().url.toString
+
+    def category(section: String, category: String): String =
+      controllers.ExercisesController.category(section, category).url.toString
+
+    def evaluate(section: String, category: String) =
+      controllers.ExercisesController.evaluate(section, category).url.toString
   }
 
 }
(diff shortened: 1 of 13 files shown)


scala-exercises-client/src/main/scala/config/Routes.scala
```diff
@@ -2,13 +2,9 @@ package config
...
-  object Users {
-    val base = "/users"
-    def all = base + "/all"
-    def create = base + "/create"
-    def update(id: Long) = base + s"/update/$id"
-    def delete(id: Long) = base + s"/delete/$id"
-    def clear = base + "/clear"
+  object Exercises {
+    val base = "/exercises/sections"
+    def evaluate(section: String, category: String) = base + s"/$section/$category/evaluate"
```
raulraja: Can't we use the Play's reverse routes here instead of recreating a Routes file manually?

scala-exercises-client/src/main/scala/scripts/ExercisesJS.scala
```diff
@@ -1,17 +1,83 @@
...
+
+  def evalExercise(exercise: JQuery) = {
+    val section = exercise.attr("data-section")
+    val category = exercise.attr("data-category")
+    val method = exercise.attr("data-exercise")
+    val args = exercise.find("input").toArray().map(e ⇒ $(e.asInstanceOf[dom.Element]).value.toString).toSeq
+    val request = EvaluationRequest(section, category, method, args)
+
+    Ajax.postAsJson(Routes.Exercises.evaluate(section, category), write(request)).map { r ⇒
+      if (r.responseText.contains("Evaluation succeded")) runSucceded(exercise, request)
```
raulraja: This should use http response codes and not string comparison

scala-exercises-client/src/main/scala/scripts/ExercisesJS.scala
```diff
@@ -1,17 +1,83 @@
...
+
+  def evalAnswered(exercise: JQuery) = {
+    val inputs = exercise.find("input")
+    val filledInputs = inputs.filter({ (index: Int, e: dom.Element) ⇒ isAnswered(e) })
+
+    if (filledInputs.length == inputs.length) evalExercise(exercise)
+  }
+
+  def evalExercise(exercise: JQuery) = {
+    val section = exercise.attr("data-section")
```
raulraja: Can these be processed in a monadic context such as `Option` such as it won't fail if param is found?

raulraja: Also these are side effecting when all these ops are processed in the async context. Doesn't Ajax provide of a monadic way of processing these ops?

raulraja: We can use an `Applicative` builder to validate all of the inputs are valid then run the ajax request

scala-exercises-client/src/main/scala/config/Routes.scala
dialelo: As far as I know it'd involve generating a [javascript router](https://www.playframework.com/documentation/2.3.0/ScalaJavascriptRouting) and using it from Scala.js. Unfortunately I haven't found a cleaner way to use it directly from client-side Scala. What's your opinion on this?

raulraja: Once the router is generated it can be invoked by Scala.js like any other JS object. The router will live in a JS object like so `jsRoutes.controllers.Users.get(someId)`. This will ensure routes are automatically kept up to date should any resource url change down the road.

rafaparadela:
Implemented in other PR #108

PR #98, opened 2016-01-05
Fix modulo

CraftThatBlock:
All values (2, 4, 6) give 0 when modulo'd by 2.


CraftThatBlock:
Whoops. I'm tired. I really should get sleep. Ignore this, I just had a brain fart.

PR #100, opened 2016-01-09
You can use `--`

cblp:


dialelo:
The symbolic alias for `diff` is `&~`. `--` is more general but it requires knowledge of the `Traversable` trait, which is introduced later in the exercises. From the Scala set docs:

>The alphabetic versions are intersect, union, and diff, whereas the symbolic versions are &, |, and &~. In fact, the ++ that Set inherits from Traversable can be seen as yet another alias of union or |, except that ++ takes a Traversable argument whereas union and | take sets.

I think the current text is fine but if you can argue that `--` is better than `&~` and won't confuse newcomers I'm happy to merge this change.

cblp:
Thank you.

PR #101, opened 2016-01-10
You don't have to specify type of the argument multiple times

cblp:
Once is enough.


dialelo:
Good point, thanks a lot.

PR #122, opened 2016-01-13
Multi build w/ plugin

andyscott:
Reorganize the project to support building a plugin in the same git repository.


raulraja:
LGTM. @dialelo and I rebased the branch to solve some conflicts with the latest merges. I think we didn't break anything but feel free to take a look in case we did. I'm merging this now so other work can proceed based on this new structure. Thanks and good job for the impressive work you are doing :+1:

PR #124, opened 2016-01-14
Syntax highlighting

dialelo:


dialelo:
Will update when we merge #134

PR #128, opened 2016-01-14
Various cleanup for recent changes/merges to version-2

andyscott:
This should be ready soon. I need to incrementally push my changes to verify that each one runs properly on the CI server (due to the weird build inconsistency we are getting w/ scalariform dependencies).


andyscott:
@rafaparadela @raulraja, what are your thoughts on dropping the `scala-exercises-` prefix for the project directories in `site`? Similarly, drop the `scalaExercises` prefix for the project names in the build file.

raulraja:
LGTM Re: name prefix, yeah they are redundant, let's rename them.

andyscott:
Cool. I will merge this after the build passes.

PR #130, opened 2016-01-16
Provide a description for getOrElse

nguyent:


app/json/options.json
```diff
@@ -19,7 +19,7 @@
       "postparagraph": ""
     },
     {
-      "preparagraph": "Provide a default value for `None`:",
+      "preparagraph": "Using getOrElse we can provide a default value ("No value") when the optional argument (`None`) does not exist:",
```
dialelo: I like the change, could you add backticks around `getOrElse` so it's shown in monospaced fonts? This way we stress the fact that is a method.

nguyent: @dialelo Done!

dialelo:
Now is much clearer for newcomers, thanks for your time!

PR #131, opened 2016-01-17
typo fix

cblp:


app/json/enumerations.json
```diff
@@ -55,7 +55,7 @@
     },
     {
       "preparagraph": "You can extend the `Enumeration` by extending the `Val` class.",
```
dialelo: Good catch! We should change Val to Value in the preparagraph too, care to make the change? Thanks a lot.

cblp:
Now ok?

dialelo:
Excellent, thanks a lot for your time!

PR #133, opened 2016-01-17
Options - Reorder post-paragraph to avoid confusion

javicg:
The type of "result1" indicated in the text refers to a previous exercise.
The correct type of "result1" in the last exercise is "Int".


dialelo:
Good catch, thanks!

PR #134, opened 2016-01-18
Organize the client-side logic around actions

dialelo:
This PR organizes the client-side program around first-class
actions. Actions represent what can happen in the system, and
they can have different outcomes. We may want to update the state
of the world responding to certain action, update the UI as a side
effect, or perform an effect (for example AJAX communication) that
may result in further actions.

# State

I've given a explicit type to the state of the application, for now
it's just a type alias. At any given time, the program has a certain
state. Whenever an action occurs, a reducing function with the signature
`(s: State, a: Action): State` yields the new state.

Since the program state changes over time, it is stored in a reactive
variable.

# UI

Since the UI is rendered by the server and the client updates it to
reflect the current state, I've thought that we can have a function
with the signature `(s: State, a: Action): IO[Unit]` that will return
an IO action that updates the DOM responding to the different actions.

# Effects

Effects such as AJAX communication can also be triggered responding to
actions. For example, we respond to the `CompileExercise` action by
performing an AJAX call to the server. Since these effects can have
different outcomes and we want to give feedback to the user, a function
with the `(s: State, a: Action): Future[Option[Action]]` signature
is used to create effects.

When an effect's response contains a `Some` with an action, we interpret
the action's state changes, UI update and further effects. For example, the
`CompileExercise` effect may trigger a `CompilationSucceeded` or
`CompilationFailed` action depending on the response of the server.

# Putting it all together

Since the UI is not declarative (not derived from program state) and
the server renders it, my approach to program bootstrap has been:

 - Load initial state of the application
 - Reflect the initial state in the UI
 - Listen to user input and convert it into actions
 - Interpret actions
   + Update state
   + Perform effects
   + Update UI

The plumbing is done using a couple reactive variables for the parts
of the program that change over time: the state and the stream of
actions.

### DIFF
--- a/site/client/src/main/scala/Model.scala
+++ b/site/client/src/main/scala/Model.scala
@@ -40,8 +40,10 @@ object Exercises {
   def updateByMethod(s: State, method: String, args: Seq[String]): State =
     applyByMethod(s, method, _.copy(arguments = args))
 
-  def evaluate(s: State, method: String): State =
-    applyByMethod(s, method, _.copy(state = Evaluating))
+  def evaluate(s: State, method: String): State = findByMethod(s, method) match {
+    case Some(exercise) if exercise.canBeCompiled ⇒ applyByMethod(s, method, _.copy(state = Evaluating))
+    case _                                        ⇒ s
+  }
 
   def setAsSolved(s: State, method: String): State =
     applyByMethod(s, method, _.copy(state = Solved))
--- a/site/client/src/main/scala/Model.scala
+++ b/site/client/src/main/scala/Model.scala
@@ -1,22 +1,26 @@
 package model
 
-case class ClientExercise(
-    method:    String,
-    arguments: Seq[String] = Nil
-) {
-
-  def isFilled: Boolean = !arguments.exists(_.isEmpty) && arguments.nonEmpty
-
-  def isSolved: Boolean = false
-
-  def canBeCompiled: Boolean = isFilled && !isSolved
-}
-
 sealed trait ExerciseState
 case object Unsolved extends ExerciseState
 case object Evaluating extends ExerciseState
 case object Errored extends ExerciseState
 case object Solved extends ExerciseState
+
+case class ClientExercise(
+  method:    String,
+  arguments: Seq[String] = Nil,
+  state: ExerciseState = Unsolved
+) {
+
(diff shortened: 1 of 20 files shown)


site/client/src/main/scala/Model.scala
```diff
@@ -0,0 +1,32 @@
...
+) {
+
+  def isFilled: Boolean = !arguments.exists(_.isEmpty) && arguments.nonEmpty
+
+  def isSolved: Boolean = false
+
+  def canBeCompiled: Boolean = isFilled && !isSolved
+}
+
+class ExerciseState extends Enumeration {
```
raulraja: We should try to avoid using enumerations where possible as they may be the source of many issues down the road.
A better type for this case would be using an ADT withc ase objects.
```scala
sealed trait ExerciseState
case object Unsolved extends ExerciseState
...
```

raulraja:
LGTM This is awesome. Great work.
Let's try to get rid of the enums though.
LGTM

dialelo:
I added a few tests for the state management logic, both the primitives in the `model` package and the action interpretation. I'm building on top of this so it'd be nice to have it merged, take a look whenever you can @rafaparadela.

rafaparadela:
I just can say:

site/client/src/main/scala/utils/DomHandler.scala
```diff
@@ -18,64 +20,105 @@ object DomHandler {
...
+  def disableCompileButton(e: HTMLElement): IO[Unit] = io {
+    $(e).find(".compile button").prop("disabled", true)
+  }
+
+  /** Given a method name, return the node corresponds to such method.
+    */
+  def nodeByMethod(method: String): IO[Option[HTMLElement]] = for {
+    exercises ← allExercises
+  } yield exercises.find(getMethodAttr(_) == method)
+
```
rafaparadela: You has re-implemented a method that already existed:

[Line 113](https://github.com/47deg/scala-exercises/pull/134/files#diff-e1b555c5df32d77153fd2dda3d1b98e8R113):
```
def findExerciseByMethod(method: String): IO[Option[HTMLElement]] = for {
    exercises ← allExercises
  } yield exercises.find(methodName(_) == Option(method))
```

dialelo:
It's ready for merge :shipit:

rafaparadela:
Come on dude... ship this :boat:

PR #136, opened 2016-01-22
Enriched state-effects-ui

rafaparadela:
Some of the changes included in this PR are:

- The `onKeyUp` event expects now two different effects; the second one for enter-key pressed, which triggers an action to compile the exercise.
- Added the condition to compile just if the exercise is filled.
- Added visual effects for user interaction as: colours of the inputs, visibility of the compile button, opacity of the ribbon,  height of the compiling message wrapper, etc.
- Has been changed the logic to update the UI, based on the class attribute of the exercise node.

@dialelo @raulraja Could you review please? Thank you.

### DIFF
--- a/site/client/src/main/scala/Ui.scala
+++ b/site/client/src/main/scala/Ui.scala
@@ -8,6 +8,14 @@ import model.Exercises._
 import utils.DomHandler._
 
 object UI {
+
+  sealed trait ExerciseStyle { def className: String }
+  case object Unsolved extends ExerciseStyle { val className = "unsolved" }
+  case object Filled extends ExerciseStyle { val className = "filled" }
+  case object Evaluating extends ExerciseStyle { val className = "evaluating" }
+  case object Errored extends ExerciseStyle { val className = "errored" }
+  case object Solved extends ExerciseStyle { val className = "solved" }
+
   def noop: IO[Unit] = io {}
 
   def update(s: State, a: Action): IO[Unit] = a match {
@@ -22,9 +30,9 @@ object UI {
     Exercises.findByMethod(s, method) match {
       case Some(exercise) ⇒ {
         if (exercise.isFilled)
-          setClassToExercise(method, "filled")
+          setClassToExercise(method, Filled.className)
         else
-          setClassToExercise(method, "unsolved")
+          setClassToExercise(method, Unsolved.className)
       }
       case _ ⇒ noop
     }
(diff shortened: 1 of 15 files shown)


site/client/src/main/scala/Ui.scala
```diff
@@ -5,54 +5,49 @@ import IO._
...
     case _                            ⇒ noop
   }
 
-  def toggleCompileButton(s: State, method: String): IO[Unit] = {
+  def toggleExerciseClass(s: State, method: String): IO[Unit] = {
     Exercises.findByMethod(s, method) match {
       case Some(exercise) ⇒ {
         if (exercise.isFilled)
-          enableCompileButton(method)
+          setClassToExercise(method, "filled")
```
raulraja: We should make these constants

site/client/src/main/scala/Ui.scala
```diff
@@ -5,54 +5,49 @@ import IO._
...
   }
 
-  def enableCompileButton(method: String): IO[Unit] = {
-    DomHandler.nodeByMethod(method)
-      .map(s ⇒ s.foreach(
-        (node ⇒ DomHandler.enableCompileButton(node).unsafePerformIO())
-      ))
-  }
+  def setClassToExercise(method: String, style: String): IO[Unit] = findExerciseByMethod(method)
+    .map(_.foreach(setClass(_, style).unsafePerformIO()))
```
raulraja: Why do we need to force execution at this point?

site/client/src/main/scala/utils/DomHandler.scala
```diff
@@ -20,39 +21,45 @@ object DomHandler {
...
+  ): IO[Unit] = io {
+    $(input).keyup((e: dom.KeyboardEvent) ⇒ {
       (for {
         _ ← OptionT(setInputWidth(input) map (_.some))
         methodName ← OptionT(io(methodParent(input)))
         exercise ← OptionT(findExerciseByMethod(methodName))
         inputsValues = getInputsValues(exercise)
-        _ ← OptionT(onkeyup(methodName, inputsValues) map (_.some))
+        _ ← {
+          if (e.keyCode == KeyCode.enter)
```
raulraja: We can wrap the whole if in optiont

site/client/src/main/scala/Ui.scala
dialelo: Since `setClass` performs mutations on the DOM it returns an `IO[Unit]` and we need to execute it, I'm not entirely happy with this either but I wanted to reflect the side effects in the type signature.

site/server/app/views/templates/library/exercise.scala.html
```diff
@@ -11,10 +11,11 @@
       </div>
 
       <div class="compile">
-        <button type="button" class="btn btn-default btn-xs" disabled>Compile</button>
```
dialelo: I set the button to `disabled` to prevent the user trying to evaluate unfinished exercises, do we want to have it enabled even when the exercises can not be compiled yet?

rafaparadela: The initial status of the button is `opacity = 0`.

rafaparadela:
@dialelo All the comments have been addressed. Is good to merge?

PR #141, opened 2016-01-25
Coerce only "true" and "false" strings to booleans

### LINKED ISSUE #137: Boolean coercion uses `false` as the default case and it shouldn't
When coercing user-submitted inputs to boolean (for example in the "Literal Booleans" exercise), strings that are converted to boolean will be `false` by default unless they are the text `"true"`. Coercing a string to a boolean should ensure that the input is either `"true"` or `"false"` and fail otherwise.

This means that the inputs that are expected to be `false` will be considered as good even when containing bogus strings that aren't `"true"`.

dialelo:
Closes #137

### DIFF
--- a/site/build.sbt
+++ b/site/build.sbt
@@ -65,7 +65,8 @@ lazy val server = (project in file("server"))
       "org.scalaz" %% "scalaz-concurrent" % "7.2.0") ++
     testlibs(
       specs2,
-      "org.typelevel" %% "scalaz-specs2" % "0.3.0")
+      "org.typelevel" %% "scalaz-specs2" % "0.3.0",
+      "org.scalacheck" %% "scalacheck" % "1.12.5")
   ))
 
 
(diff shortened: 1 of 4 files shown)


dialelo:
Just rebased on top of `version-2`and this merges cleanly, @raulraja should we make any modifications here?

raulraja:
This is great but it's a great example of something we can test with ScalaCheck for all strings not in that set. Let's add it if you also find it appropriate so there is some good examples of the types of tests that we expect in future contributions.

dialelo:
Good point!

dialelo:
@raulraja check it out please.

raulraja:
LGTM! Would love to have @noelmarkham look at it in case he can suggest any improvements but this is great! 

LGTM

noelmarkham:
LGTM - I like it :+1:

dialelo:
\o/

Issue #113, opened 2016-01-13
Add GitHub user authentication


rafaparadela:
The current Github integration and login came from an example which is far from clean. We need to integrate the login ops as its own algebra in the context Free as we have done with other Biz logic ops.

dialelo:
Note that currently GH authentication can throw runtime exceptions which we should avoid. We have to use Play applicative builders for validating data.

rafaparadela:
When addressing this issue we can considerate to review the permissions we are requesting (we only need github loing and email).

andyscott:
Is this done now? @rafaparadela @raulraja

raulraja:
@andyscott It's half way implemented. We have the algebra, ops and doobie stuff all wired but `com.fortysevendeg.exercises.utils.OAuth2` still needs cleanup

dialelo:
I'm going to clean that up so we can close this, assigning the issue to me.

PR #143, opened 2016-01-28
User persistence with Doobie

dialelo:
~~Not ready for merge but opening for feedback, feel free to modify the list below~~:

- [x] Use Doobie instead of Slick for DB access
- [x] Dependency Injection with implicits
- [x] User operations algebra
- [x] User operations production interpreter
- [x] Better type signatures for user store & service
- [x] Better schema for users
 + [x] Evolution for the new schema 

Is not yet ready but any comments welcome, @raulraja also suggested lifting `ConnectionIO` to cats' Free monad which we can do in the future. Let me know your thoughts @rafaparadela @juanpedromoreno @andyscott.


site/build.sbt
```diff
@@ -62,7 +62,8 @@ lazy val server = (project in file("server"))
       "org.scala-lang" % "scala-compiler" % scalaVersion,
       "org.clapper" %% "classutil" % "1.0.5",
       "com.toddfast.typeconverter" % "typeconverter" % "1.0",
-      "org.scalaz" %% "scalaz-concurrent" % "7.2.0") ++
+      "org.scalaz" %% "scalaz-concurrent" % "7.1.2",
```
juanpedromoreno: Maybe we could unify somehow the scalaz dependency version, since we are using different ones along the different modules: 7.1.2, 7.1.4, 7.2.0...

site/server/app/com/fortysevendeg/exercises/ApplicationLoader.scala
```diff
@@ -26,15 +33,15 @@ class Components(context: Context)
...
     with HikariCPComponents {
 
-  lazy val router = new Routes(httpErrorHandler, applicationController, userController, exercisesController, assets, OAuthController)
+  val dataSource = dbApi.database("default").dataSource
+  implicit val transactor: Transactor[Task] = DataSourceTransactor[Task](dataSource)
+
+  val applicationController = new ApplicationController
+  val exercisesController = new ExercisesController
+  val userController = new UserController
+  val OAuthController = new OAuth2Controller
```
juanpedromoreno: IMHO `val` better if they are `camelCase` ;)

oauthController perhaps?

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -18,13 +18,14 @@ import play.api.mvc._
...
+import com.fortysevendeg.exercises.services.interpreters.ProdInterpreters._
+
 import scalaz.concurrent.Task
 import scalaz.{ \/-, -\/ }
 
 import scala.concurrent.Future
 
-class ApplicationController(userService: UserServices)(implicit exerciseOps: ExerciseOps[ExercisesApp]) extends Controller {
-
+class ApplicationController(implicit exerciseOps: ExerciseOps[ExercisesApp], userService: UserServices) extends Controller {
```
juanpedromoreno: Line limit exceeded :)

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -35,8 +36,8 @@ class ApplicationController(userService: UserServices)(implicit exerciseOps: Exe
     exerciseOps.getLibraries.runTask match {
       case \/-(libraries) ⇒
         request.session.get("oauth-token") map { token ⇒
-          val user = userService.getUserByLogin(GetUserByLoginRequest(login = request.session.get("user").getOrElse("")))
-          user.map(response ⇒ Ok(views.html.templates.home.index(user = response.user, libraries = libraries)))
+          val user = userService.getUserByLogin(request.session.get("user").getOrElse(""))
+          Future.successful(Ok(views.html.templates.home.index(user = user, libraries = libraries)))
```
juanpedromoreno: Minor question, if the user hasn't been found in the session... should we return `Future.successful`?

site/server/app/com/fortysevendeg/exercises/controllers/ExercisesController.scala
```diff
@@ -10,10 +10,11 @@ import scala.concurrent.Future
 import scalaz.{ -\/, \/, \/- }
 
 import com.fortysevendeg.exercises.services._
-import com.fortysevendeg.shared.free.ExerciseOps
 import com.fortysevendeg.exercises.app._
+import com.fortysevendeg.shared.free.ExerciseOps
+import com.fortysevendeg.exercises.services.interpreters.ProdInterpreters._
 
-class ExercisesController(implicit exerciseOps: ExerciseOps[ExercisesApp]) extends Controller with JsonFormats {
+class ExercisesController(implicit exerciseOps: ExerciseOps[ExercisesApp], userServices: UserServices) extends Controller with JsonFormats {
```
juanpedromoreno: Same here (long line ;) )

site/server/app/com/fortysevendeg/exercises/services/UserServices.scala
```diff
@@ -1,71 +1,79 @@
...
+  def createUser(
+    login:       String,
+    name:        String,
+    github_id:   String,
+    picture_url: String,
+    github_url:  String,
+    email:       String
+  ): Throwable Xor User = {
+    val user = User(None, login, name, github_id, picture_url, github_url, email)
+    Xor.fromOption(userStore.create(user), new Exception("Couldn't create user"))
```
juanpedromoreno: I think we don't want generic exceptions. We could wrap this exception in a more specific one.

juanpedromoreno:
Minor format comments added. I like this PR!! Good job!

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
dialelo: I think so, we can have anonymous users navigate the site with no problem. This won't allow exercise evaluation but the content is rendered so web crawlers can index it.

site/server/app/com/fortysevendeg/exercises/services/UserServices.scala
dialelo: Definitely, I'm going to improve the error handling and type signatures.

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
juanpedromoreno: Makes sense :+1:

dialelo:
I've made a few changes here, namely:

- Change APIs to avoid common pitfalls
 + Don't accept `User` as a parameter to update functions since its id might be `None`
 + Don't accept `User` as a parameter to create functions since ID's are provided by the DB
- Separate DB access in a couple of layers
 + `com.fortysevendeg.exercises.models.queries.Queries` constructs queries, which are typechecked in tests
 + `com.fortysevendeg.exercises.models.UserDoobieStore` constructs DB operations yielding `ConnectionIO` for composability, it doesn't implicitly require a transactor and doesn't run queries
 + `com.fortysevendeg.exercises.services.UserServiceImpl` takes an implicit transactor and runs operations in `UserDoobieStore`

I'd like to improve error handling but is difficult to do it in a DB-agnostic way. Since we are switching to Postgres soon my idea is to take advantage of Doobie's [Postgres error handling facilities](http://tpolecat.github.io/doobie-0.2.3/13-Extensions-PostgreSQL.html) for error handling to improve it, but that will be another PR.

site/build.sbt
```diff
@@ -27,6 +27,8 @@ lazy val exerciseSettings: Seq[Setting[_]] =
   )
 
 lazy val clients = Seq(client)
+lazy val doobieVersion = "0.2.3"
+lazy val scalazVersion = "7.1.4"
```
raulraja: scalaz-concurrent already brings in scalaz and so does doobie. We should try to align versions if they aren't already.

site/server/app/com/fortysevendeg/exercises/controllers/UserController.scala
```diff
@@ -6,24 +6,32 @@ import play.api.libs.json.Reads._
...
+import com.fortysevendeg.exercises.services._
+import com.fortysevendeg.exercises.services.free.UserOps
+import com.fortysevendeg.exercises.app._
+import com.fortysevendeg.exercises.services.interpreters.ProdInterpreters._
 
 import scala.concurrent.Future
+import scalaz.{ -\/, \/, \/- }
 
-class UserController(userStore: UserStore) extends Controller {
+class UserController(implicit userOps: UserOps[ExercisesApp], userServices: UserServices) extends Controller {
```
juanpedromoreno: Is `userServices` needed in this context?

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -35,8 +40,8 @@ class ApplicationController(userService: UserServices)(implicit exerciseOps: Exe
     exerciseOps.getLibraries.runTask match {
       case \/-(libraries) ⇒
         request.session.get("oauth-token") map { token ⇒
-          val user = userService.getUserByLogin(GetUserByLoginRequest(login = request.session.get("user").getOrElse("")))
-          user.map(response ⇒ Ok(views.html.templates.home.index(user = response.user, libraries = libraries)))
+          val user = userService.getUserByLogin(request.session.get("user").getOrElse(""))
```
raulraja: If `userService.getUserByLogin` returns an `Option` it means that it may be blocking. In which case the default executor pool may not be appropriate. This should all be in the context of `Free` and monadically computed along with `getLibraries` above so it can be lifted to the appropriate interpreter target monad.

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -20,81 +20,99 @@ trait UserStore {
...
+  ): ConnectionIO[Option[User]] = for {
+    _ ← Queries.insert(login, name, github_id, picture_url, github_url, email).run
+    user ← getByLogin(login)
+  } yield user
+
+  def deleteQuery(login: String): ConnectionIO[Boolean] = {
+    Queries.byLogin(login).option.flatMap({
+      case None ⇒ sql"select false".query[Boolean].unique
+      case Some(u) ⇒ Queries.update(
+        u.id.get,
```
juanpedromoreno: We shouldn't use `Option.get` never ;)

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -20,81 +20,99 @@ trait UserStore {
...
+    require(user.id.isDefined, "Can only update existing users")
 
-  override def delete(ids: Long*): Future[Boolean] = {
-    Future.sequence {
-      for {
-        id ← ids
-      } yield { db.run(users.filter(_.id === id).delete).map(_ == 1) }
-    }.map(!_.exists(i ⇒ !i))
+    Queries.update(
+      user.id.get,
```
juanpedromoreno: Same here.

site/server/app/com/fortysevendeg/exercises/controllers/UserController.scala
dialelo: Not at all, good catch!

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -20,81 +20,99 @@ trait UserStore {
...
-    def name = column[String]("name")
-    def github_id = column[String]("github_id")
-    def picture_url = column[String]("picture_url")
-    def github_url = column[String]("github_url")
-    def email = column[String]("email")
-    def * = (id, login, name, github_id, picture_url, github_url, email) <> (User.tupled, User.unapply)
+object UserDoobieStore extends UserStore {
+  def all: ConnectionIO[List[User]] = {
+    Queries.all.list
   }
```
raulraja: No need for {} here

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -20,81 +20,99 @@ trait UserStore {
...
 
-  val users = TableQuery[Users]
-
-  override def all: Future[Seq[User]] = {
-    db.run(users.sortBy(_.id.desc).result)
+  def getByLogin(login: String): ConnectionIO[Option[User]] = {
+    Queries.byLogin(login).option
   }
```
raulraja: See comment above regarding {}

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -20,81 +20,99 @@ trait UserStore {
...
+      user.name,
+      user.github_id,
+      user.picture_url,
+      user.github_url,
+      user.email
+    ).run
   }
 
+  def update(
+    id:          Int,
```
raulraja: This should take a container instead of so many parameters. In general we favor single arg functions vs multiple ones.

site/server/app/com/fortysevendeg/exercises/services/interpreters/interpreters.scala
```diff
@@ -25,15 +30,42 @@ object ProdInterpreters extends InterpreterInstances {
...
-  implicit def userOpsInterpreter: UserOp ~> Task = new (UserOp ~> Task) {
-
-    def apply[A](fa: UserOp[A]): Task[A] = ???
-
+  def userOpsInterpreter: UserOp ~> Task = new (UserOp ~> Task) {
+
+    def apply[A](fa: UserOp[A]): Task[A] = fa match {
+      case GetUsers()            ⇒ Task.delay(userService.all)
+      case GetUserByLogin(login) ⇒ Task.delay(userService.getUserByLogin(login))
+      case CreateUser(
```
raulraja: CreateUser should have been passed straight to the service as it may be decomposed there and applied to the query if needed via unapply

site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
```diff
@@ -24,10 +25,12 @@ object OAuth2 {
   }
 }
 
-class OAuth2Controller(ws: WSClient, userService: UserServices) extends Controller {
+class OAuth2Controller(implicit userService: UserServices) extends Controller {
 
   import OAuth2._
 
+  val ws = NingWSClient()
```
raulraja: We need to look into the NingWSClient lifecycle in case it needs to be closed to avoid leaks like some clients have in the past.

site/server/conf/evolutions/default/2.sql
```diff
@@ -0,0 +1,16 @@
+# --- !Ups
+
+DROP TABLE users;
+CREATE TABLE users (
+    id integer primary key,
+    login varchar(255) unique not null,
+    name varchar(255) not null,
+    github_id varchar(255) not null,
```
raulraja: We should try to avoid these _ leaking through Scala code and SQL if possible since it is not common practice in Scala to include such name conventions.

site/server/test/Database.scala
```diff
@@ -0,0 +1,32 @@
...
+object TestDatabase {
+  val testDriver = "org.h2.Driver"
+  def testUrl = {
+    val dbFile = File.createTempFile("temporary-db", ".tmp").getAbsolutePath()
+    s"jdbc:h2:file:$dbFile;DATABASE_TO_UPPER=false"
+  }
+  val testUsername = "sa"
+  val testPassword = ""
+
+  def db: Database = {
```
raulraja: The name `db` it's misleading since it applies effects in applyEvolutions. Perhaps something like `updateDB()`

site/server/test/UserStoreSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+  val transactor: Transactor[Task] = TestDatabase.transactor
+  import transactor.yolo._
+
+  def before = UserDoobieStore.deleteAll.quick.run
+
+  "UserDoobieStore" should {
+    "create a new User and retrieve it afterwards" in {
+      val aUser = User(None, "47deg", "47 degrees", "47deg", "http://placekitten.com/50/50", "http://github.com/47deg", "hi@47deg.com")
+      val storedUser = UserDoobieStore.create(
+        aUser.login,
```
raulraja: If User and CreateUser had similar shapes they could be automatically copied over

site/shared/src/main/scala/shared/User.scala
```diff
@@ -3,7 +3,7 @@ package shared
 // TODO: is any of this actually used/needed by the client js code?
 
 case class User(
-  id:          Option[Long],
+  id:          Option[Int],
```
raulraja: Why this change from Long to Int?

site/server/conf/evolutions/default/3.sql
```diff
@@ -0,0 +1,16 @@
+# --- !Ups
```
juanpedromoreno: If we are working in a PR in isolation maybe it makes sense to squash all **our** SQL changes in a single evolution. Do you agree?

site/server/test/UserStoreSpec.scala
```diff
@@ -0,0 +1,103 @@
...
+
+class UserStoreSpec extends Specification with BeforeEach {
+  val transactor: Transactor[Task] = TestDatabase.transactor
+  import transactor.yolo._
+
+  def before = UserDoobieStore.deleteAll.quick.run
+
+  "UserDoobieStore" should {
+    "create a new User and retrieve it afterwards" in {
+      val aUser = User(None, "47deg", "47 degrees", "47deg", "http://placekitten.com/50/50", "http://github.com/47deg", "hi@47deg.com")
```
juanpedromoreno: Same line in several places in this file. Could we extract this kind of code in factory method? We could receive all the parameters but with the less important ones with default values?

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
dialelo: What do you think about having an ADT with two constructors: `NewUser` and `User` instead of making the ID an `Option`? `NewUser` would have all the fields needed to create a user and `User`'s ID wouldn't be an `Option` anymore.

site/shared/src/main/scala/shared/User.scala
dialelo: My bad, I'm going to change the `id`'s type in the database so we can map it to a `Long` in our code and have more available ID's than using integers.

site/server/test/UserStoreSpec.scala
dialelo: Great point.

site/server/conf/evolutions/default/2.sql
juanpedromoreno: Agreed with @raulraja 

Particularly, I also like table names in UpperCase and using non-plural entity names:

For instance:

`CREATE TABLE User (... , githubId varchar(255), ...)`

This naming convention prevents things like this in our persistence model:

`val user1 = Users(1, "", "" , .. )`

But this a personal preference. What do you think guys @raulraja  @dialelo ?

dialelo: :+1:

site/server/conf/evolutions/default/3.sql
dialelo: Definitely, I'm going to squash all the evolutions into one with the final schema.

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
raulraja: Not saying we should do any of this now but if the UserService is not already lifting to Free or using Free Algebras we should refactor it or have a ticket to do so.

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
raulraja: A for comprehension would be better suited for this case perhaps

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
dialelo: I'll try to address it as part of the proposed changes; if I see that is too involved I'll open an issue and work on it on a separate PR for making review easier.

site/build.sbt
dialelo: The "7.1.4" version is used, both for `scalaz-core` and `scalaz-concurrent`, ignoring the transitive dependency for doobie.

dialelo:
I've addressed the comments in the PR and added a DB result algebra and implicit conversion that should allow us to run doobie transactions in our free for comprehensions. The `ExerciseServiceSpec` test fails but I can address that in a separate PR. 

Please take a look @raulraja  @juanpedromoreno @rafaparadela @andyscott :sparkles:

site/server/app/com/fortysevendeg/exercises/app.scala
```diff
@@ -1,27 +1,12 @@
...
-
-  implicit class FreeOps[A](f: Free[ExercisesApp, A]) {
-
-    /** Run this Free structure folding it's ops to an effect capturing task */
-    def runTask: Throwable \/ A =
-      f.foldMap(interpreters).unsafePerformSyncAttempt
-
-  }
-
+  type ExercisesAndUserOps[A] = Coproduct[ExerciseOp, UserOp, A]
```
raulraja: I mentioned something similar in the PR from @juanpedromoreno. I don't believe the intermediate `Coproduct` aliases need to be named like these because they are not going to be used in isolation or composed in other context. Since they just represent aliases in the `Coproduct` tree constructions they may as well be named something like `C01`. If you disagree is fine, just a thought, since the only thing I need to know about those type aliases is in which order they are getting composed. It's similar to the use of the lambda name in type lambdas, this is just Scala being annoying in the lack of support for currying type constructors.

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -1,29 +1,34 @@
 package com.fortysevendeg.exercises.controllers
 
+import scala.language.postfixOps
```
raulraja: We should add all these flags in the build.sbt as compiler flags. We will use infix notation and higher kinds elsewhere and there is no need to have these unnecesary advanced lang imports everywhere

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -1,29 +1,34 @@
...
 import scalaz.{ \/-, -\/ }
 
 import scala.concurrent.Future
 
-class ApplicationController(userService: UserServices)(implicit exerciseOps: ExerciseOps[ExercisesApp]) extends Controller {
+class ApplicationController(
+    implicit
+    exerciseOps: ExerciseOps[ExercisesApp],
+    userOps:     UserOps[ExercisesApp],
+    T:           Transactor[Task]
```
raulraja: We should have cohesive naming for implicit evidences across the board.

site/server/app/com/fortysevendeg/exercises/controllers/UserController.scala
```diff
@@ -1,29 +1,44 @@
...
+    userOps: UserOps[ExercisesApp],
+    T:       Transactor[Task]
+) extends Controller {
 
   implicit val jsonReader = (__ \ 'github).read[String](minLength[String](2))
 
-  def index = Action { implicit request ⇒
-    Redirect("/")
+  def all = Action { implicit request ⇒
+    userOps.getUsers runTask match {
```
raulraja: We should consider `runTask` to return an `Xor[A]` since we use cats elsewhere

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -1,100 +1,78 @@
...
-        picture_url = picture_url,
-        github_url = github_url,
-        email = email
-      )
-    }
-  }
+  def create(user: UserCreation.Request): ConnectionIO[UserCreation.Response] = for {
+    _ ← Queries.insert(user).run
+    user ← getByLogin(user.login)
+  } yield if (user.isDefined) Xor.Right(user.get) else Xor.Left(UserCreation.DuplicateName)
```
raulraja: A `fold` or `pattern matching` is better suited for this case than manual runtime boolean check on presence / absence. It's literally safe the way it is but we don't want to encourage use of `Option.get` or any unsafe state accessors anywhere

site/server/app/com/fortysevendeg/exercises/models/UserModel.scala
```diff
@@ -1,100 +1,78 @@
...
-      val q = for {
-        u ← users if u.id === user.id
-      } yield (u.login)
-      q.update(user.login)
-    }.map(_ == 1)
-  }
+  def getOrCreate(user: UserCreation.Request): ConnectionIO[UserCreation.Response] = for {
+    maybeUser ← getByLogin(user.login)
+    theUser ← if (maybeUser.isDefined)
+      Xor.Right(maybeUser.get).point[ConnectionIO]
```
raulraja: See comment above regarding the use of unsafe methods

site/server/app/com/fortysevendeg/exercises/services/interpreters/interpreters.scala
```diff
@@ -22,20 +42,54 @@ trait Interpreters[F[_]] extends InterpreterInstances {
...
     }
-
   }
 
-  implicit def userOpsInterpreter(implicit A: Applicative[F]): UserOp ~> F = new (UserOp ~> F) {
-
-    def apply[A](fa: UserOp[A]): F[A] = ???
+  implicit def userOpsInterpreter(implicit A: ApplicativeError[F, Throwable], T: Transactor[F]): UserOp ~> F = new (UserOp ~> F) {
 
+    def apply[A](fa: UserOp[A]): F[A] = fa match {
```
raulraja: `import UserDoobieStore._`

site/server/app/com/fortysevendeg/exercises/services/interpreters/interpreters.scala
```diff
@@ -48,12 +102,24 @@ trait InterpreterInstances {
...
+  }
 }
 
 /** Production based interpreters lifting ops to the effect capturing scalaz.concurrent.Task **/
-object ProdInterpreters extends Interpreters[Task]
+object ProdInterpreters extends Interpreters[Task] {
+  implicit class FreeOps[A](f: Free[ExercisesApp, A]) {
+
+    /** Run this Free structure folding it's ops to an effect capturing task */
+    def runTask(implicit T: Transactor[Task]): Throwable \/ A = {
```
raulraja: Let's return an `Xor` here so we don't leak the Scalaz types since we are using cats elsewhere

site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
```diff
@@ -24,7 +30,11 @@ object OAuth2 {
   }
 }
 
-class OAuth2Controller(ws: WSClient, userService: UserServices) extends Controller {
+class OAuth2Controller(
+    implicit
+    transactor: Transactor[Task],
```
raulraja: This is named `T` in other places, we should change all implicit evidences to have a cohesive naming convention.

site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
```diff
@@ -72,23 +82,24 @@ class OAuth2Controller(ws: WSClient, userService: UserServices) extends Controll
 
           val login = (response.json \ "login").as[String]
           val name = (response.json \ "name").as[String]
-          val github_id = (response.json \ "id").as[Long]
-          val avatar_url = (response.json \ "avatar_url").as[String]
-          val html_url = (response.json \ "html_url").as[String]
+          val githubId = (response.json \ "id").as[Long]
```
raulraja: This is not safe but I understand we are changing this in another PR. All JSON parsing ops should be treated as an applicative builder style where a chunk is reduce to the structure or errors are reported accordingly, similar how the functional builders for JSON in play work.

site/server/test/Database.scala
```diff
@@ -0,0 +1,29 @@
...
+  val testUsername = "sa"
+  val testPassword = ""
+
+  def create: Database = {
+    Databases(
+      driver = testDriver,
+      url = testUrl,
+      config = Map("user" → testUsername, "password" → testPassword)
+    )
+  }
```
raulraja: `{...}` Braces may be ommited

site/server/test/UserQueriesSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+      "47deg",
+      "http://placekitten.com/50/50",
+      "http://github.com/47deg",
+      "hello@47deg.com"
+    )
+  ))
+  check(Queries.deleteById(47))
+  check(Queries.update(
+    User(
+      47,
```
raulraja: Use named parameters when params lists exceed 4-6 for readability

site/server/test/UserStoreSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+class UserStoreSpec extends Specification with BeforeEach {
+  val db = TestDatabase.create
+  TestDatabase.update(db)
+
+  val transactor: Transactor[Task] = TestDatabase.transactor(db)
+  import transactor.yolo._
+
+  def before = UserDoobieStore.deleteAll.quick.run
+
+  def newUser(
```
raulraja: By the way these params are applied this method may be not necessary since it's delegating all of them in order.

site/server/app/com/fortysevendeg/exercises/controllers/UserController.scala
dialelo: Done.

site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
dialelo: I added a comment in #113 to address it in another PR.

dialelo:
I've added property-based test for the user store.

site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
dialelo: In other places we needed the transactor implicit or otherwise the compiler would complain, I called it `T` because we were merely proving its existence. In this case I actually use the transactor instance so called it `transactor`, I'll rename it to `T` for consistency.

dialelo:
Since we're not using Play's integration with Slick anymore evolutions aren't applied. Any ideas on how to make evolutions run automatically (at least in DEV mode)? @raulraja @juanpedromoreno

site/server/app/com/fortysevendeg/exercises/app.scala
dialelo: Nice, I'll change it to reflect that these are intermediate coproducts that shouldn't be referred to elsewhere, they are an implementation detail.

juanpedromoreno:
@dialelo @raulraja Let's create a new ticket to run the evolutions automatically with [Flyway](https://flywaydb.org/)

dialelo:
Created #161 to address schema evolutions.

site/server/app/com/fortysevendeg/exercises/services/interpreters/interpreters.scala
```diff
@@ -48,12 +102,27 @@ trait InterpreterInstances {
...
 
+    override def handleErrorWith[A](fa: Task[A])(f: Throwable ⇒ Task[A]): Task[A] =
+      fa.handleWith({ case x ⇒ f(x) })
+  }
 }
 
 /** Production based interpreters lifting ops to the effect capturing scalaz.concurrent.Task **/
-object ProdInterpreters extends Interpreters[Task]
+object ProdInterpreters extends Interpreters[Task] {
+  def scalazToCatsDisjunction[A, B](disj: \/[A, B]): Xor[A, B] =
```
raulraja: This is cool but it could also be like...
```scala
def scalazToCatsDisjunction[A, B](disj: A \/ B): A Xor B
```

dialelo:
@raulraja just got rid of all the remaining calls to Option.get :fire:, take a look please!

raulraja:
Awesome! LGTM :+1: 
[LGTM](http://lgtm.in/i/cgfWfFNnq)

dialelo:
:sparkles: :shipit: :sparkles:

PR #158, opened 2016-02-05
Preferred default arguments

cakcora:
Previously:
howMuchCanIMake_?(30)
howMuchCanIMake_?(95,95)

This did not show that the second call was overrriding the 'amount' 34 with 95, because hours changed from 30 to 95 too. 

New version:
howMuchCanIMake_?(30)
howMuchCanIMake_?(30,95) // the answer 2850 is also updated.


dialelo:
Excellent point, thanks for your time!

PR #159, opened 2016-02-05
Added extensible interpreters for any F[_] including cats.Id

raulraja:
The following PR allows constructions of interpreters from `F[_] to G[_]` if there is implicit instance evidence of `Applicative[G]` in scope lifting all target values via `Applicative#pureEval`

@dialelo @juanpedromoreno Can you please review? Thanks!


dialelo:
:+1: Great work, LGTM!

juanpedromoreno:
:+1:

PR #162, opened 2016-02-09
Fix user store property

dialelo:
Trivial stuff, I missed it since we are not running tests in CI yet (I'm on it).


rafaparadela:
LGTM

Issue #151, opened 2016-02-02
Prepare deployment environment for version 2


dialelo:
We'll need a list of the software requirements of the Play application in order to provision the servers.

raulraja:
Initially we are planning to deploy in Heroku which supports Play 2 apps out of the box and there is already an app provisioned for it. @rafaparadela has already been given access and just needs to setup the `app.json` required to provision `sass` and the other dependencies.

rafaparadela:
We have merged this advances in `version-2` (PR #164), since we can deploy our branches in prod using the `sbt-heroku` plugin. We only need run:

```bash
sbt clean stage deployHeroku
```

However, we're not able to enable the auto deployment for PRs, since the current structure of the project is not compatible with the existing buildpacks. So I've created the ticket #176 to re-configure the structure of the modules, once we have the compiler project (and maybe others) isolated in sonatype.

PR #164, opened 2016-02-10
Switch postgres & set deployment mechanism

rafaparadela:
This PR: 

- Fixes: #149 Switch to the Postgres database for persistence
- Fixes: #151 Prepare deployment environment for version 2

@dialelo @juanpedromoreno @raulraja Could you review please?

_Notes:_
- Command to deploy: 
```bash
sbt clean stage deployHeroku
```

- Command to monitor logs: 
```bash
heroku logs --app scala-exercises --tail
```

- Prod. URL: http://scala-exercises.herokuapp.com/


raulraja:
LGTM! Let's get this in so we can setup PR review apps soon! Great stuff, thanks for doing this so quickly!

rafaparadela:
I've been trying to deploy it with a simple `git push heroku master` but it doesn't detect the `Procfile`. It seems like we have to define it in the `server` module level.

site/server/conf/evolutions/default/1.sql
```diff
@@ -1,15 +1,15 @@
...
-    id bigint primary key auto_increment,
-    login varchar(255) unique not null,
-    name varchar(255) not null,
-    githubId varchar(255) not null,
-    pictureUrl varchar(255) not null,
-    githubUrl varchar(255) not null,
-    email varchar(255) not null
+CREATE TABLE users (
+    id bigserial primary key,
+    login text UNIQUE NOT NULL,
```
dialelo: Doesn't Doobie complain about materializing `text` fields into a Scala `String`?

dialelo:
Other than the question about converting `text` fields to strings LGTM, great work!

dialelo:
From a conversation with @juanpedromoreno: We can still use the name `User` for the user table although the name has to be quoted in the queries, like the following: `SELECT * FROM "User"`

site/server/conf/evolutions/default/1.sql
rafaparadela: it doesn't. It only complains when weren't defined as 'NOT NULL'. Tests are passing successfully.

PR #172, opened 2016-02-10
Fixing evolution & evaluate ajax call

rafaparadela:
This PR resolves the issues:
- Evolution sql written as mysql style, instead of postgreslq.
- Unable to find the route to call the evaluation endpoint.

@andyscott  you should be able to have your stuff working once this PR is merged into version-2.
@juanpedromoreno  you should be able to run the ajax call to `evaluate` exercises in the client.

Note: 
- Remember to setup a postgres database called `scalaexercises`

@raulraja @dialelo Could you take a look please? thank you.

### DIFF
--- a/site/client/src/main/scala/config/Routes.scala
+++ b/site/client/src/main/scala/config/Routes.scala
@@ -3,7 +3,7 @@ package config
 import scala.scalajs.js
 
 object Routes {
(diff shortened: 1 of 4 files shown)


raulraja:
LGTM

PR #177, opened 2016-02-10
Re-enable evolutions + Use H2 for local run

andyscott:
This re-enables evolutions. Additionally:

- `sbt run` will use H2 in PostgreSQL mode
- `sbt start` (or other packaged build) will use PostgreSQL


juanpedromoreno:
LGTM.

Just a quick question. I see you are quoting the "users" table in the 1.sql evolution. It's something that I had to do yesterday, but @rafa didn't need today. So, just to double check, should the database tables be quoted? (We need to standardize that). Thanks!

andyscott:
I read somewhere that we should be quoting them. I don't think it is necessary, but I figured we should be consistent-- so I added quotes to "users".

juanpedromoreno:
:+1:

PR #178, opened 2016-02-11
Compiler updates and fixes

andyscott:
This will include various updates and fixes I'm find as I work through a larger exercise example for the content project.

### DIFF
--- a/core/sbt-exercise/src/main/scala/com/fortysevendeg/exercises/sbtexercise/ExerciseCompilerPlugin.scala
+++ b/core/sbt-exercise/src/main/scala/com/fortysevendeg/exercises/sbtexercise/ExerciseCompilerPlugin.scala
@@ -2,35 +2,36 @@
 package com.fortysevendeg.exercises
 package sbtexercise
 
+import scala.language.implicitConversions
 import scala.language.reflectiveCalls
+
+import scala.io.Codec
+import scala.util.{ Try, Success, Failure }
+import java.io.File
+import java.net.URLClassLoader
 
 import sbt.{ `package` ⇒ _, _ }
 import sbt.Keys._
 import sbt.classpath.ClasspathUtilities
 import xsbt.api.Discovery
 
-import scala.io.Codec
-import scala.util.{ Try, Success, Failure }
-
-import java.io.File
-import java.net.URLClassLoader
-
 import cats._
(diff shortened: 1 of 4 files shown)


andyscott:
For bubbling up errors, I wanted to capture both a descriptive message and/or the exception thrown. This seemed like a good time to give `Ior` a try. In this case we get an `Xor[Ior[String, Throwable], ?]`.

andyscott:
Please review and merge!

core/sbt-exercise/src/main/scala/com/fortysevendeg/exercises/sbtexercise/ExerciseCompilerPlugin.scala
```diff
@@ -176,7 +185,15 @@ object ExerciseCompilerPlugin extends AutoPlugin {
...
-    result.fold(message ⇒ throw new Exception(message), { value ⇒ value })
+    result.fold({
+      _ match {
+        case Ior.Left(message) ⇒ throw new Exception(message)
+        case Ior.Right(error)  ⇒ throw error
+        case Ior.Both(message, error) ⇒
+          log.error(message)
+          throw error
+      }
+    }, { value ⇒ value })
```
dialelo: We can just use `identity` here instead of `{ value=> value }`.

dialelo:
Other than the minor nitpick LGTM!

juanpedromoreno:
@dialelo Comment addressed

core/sbt-exercise/src/main/scala/com/fortysevendeg/exercises/sbtexercise/ExerciseCompilerPlugin.scala
```diff
@@ -176,7 +185,15 @@ object ExerciseCompilerPlugin extends AutoPlugin {
       result ← libraries.map(invokeCompiler(compiler, _)).sequenceU
```
raulraja: If it's a matter of traversing over sources and invoking the compiler and accumulating errors we could also have used `Validated` here. I'm not saying there is something wrong with this but since each compilation computation does not depend on the previous one an `Applicative` instance may be more appropriate. `Validated` comes for free with `NonEmptyList` semantics and support for the Applicative Builder syntax.

PR #179, opened 2016-02-11
Bring back exercise content

andyscott:
Begin to add back exercise content.

A few notes:
- This more clearly demonstrates the compiler and highlights some of the issues that need to be addressed.
- This needs to be merged after #178


juanpedromoreno:
LGTM

PR #181, opened 2016-02-11
Fix Save User Progress

juanpedromoreno:
@dialelo @raulraja @rafaparadela @andyscott 
Please, review at your convenience. Thanks!


site/client/src/main/scala/Api.scala
```diff
@@ -15,7 +15,8 @@ import org.scalajs.dom.ext.{ Ajax, AjaxException }
 object Client {
   def compileExercise(e: ClientExercise): Future[EvaluationResult] = {
     val url = Routes.Exercises.evaluate(e.library, e.section)
-    val request = EvaluationRequest(e.library, e.section, e.method, e.arguments)
+    //TODO: TBD version and exercise types
+    val request = EvaluationRequest(e.library, e.section, e.method, 1, "Koans", e.arguments)
```
raulraja: We should use a case object instead of a string

raulraja:
LGTM

site/client/src/main/scala/Api.scala
juanpedromoreno: Yep, we have a case object in the backend Controller. Given this code it's on the UI side, I was wondering if we can use the case object as well.

PR #184, opened 2016-02-12
Fix method body extraction

andyscott:
This is a fix for method body extractions in the compiler. I've also added some new tests.

### DIFF
--- a/core/compiler/src/test/scala/com/fortysevendeg/exercises/compiler/MethodBodyReaderSpec.scala
+++ b/core/compiler/src/test/scala/com/fortysevendeg/exercises/compiler/MethodBodyReaderSpec.scala
@@ -3,8 +3,6 @@ package compiler
 
 import org.scalatest._
 
-import java.lang.ClassLoader
-
 class MethodBodyReaderSpec extends FunSpec with Matchers with MethodBodyReaderSpecUtilities {
 
   describe("code snippet extraction") {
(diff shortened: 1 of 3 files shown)


core/compiler/src/test/scala/com/fortysevendeg/exercises/compiler/MethodBodyReaderSpec.scala
```diff
@@ -22,34 +17,99 @@ class MethodBodyReaderSpec extends FunSpec with Matchers with MethodBodyReaderSp
...
+       |  */
+       |def addOne(value: Int) = {
+       |    println("this is the first line")
+       |
+       |
+       |  println("this is the last line")
+       |}
+       """.stripMargin
+
+      println("*" * 20)
```
dialelo: We can get rid of these debugging statements.

dialelo:
Went ahead and deleted some debugging statements, LGTM!

dialelo:
After a conversation with @raulraja we both agreed that we want to keep the newlines in method bodies. They are often useful for separating chunks of code like in the following example:

```scala
import cats.data.Xor

val right: String Xor Int = Xor.right(5)
right.flatMap(x => Xor.right(x + 1)) should be(Xor.right(res0))

val left: String Xor Int = Xor.left("Something went wrong")
left.flatMap(x => Xor.right(x + 1)) should be(Xor.left(res1))
```

andyscott:
My intention was to keep them but I think I wound up removing them. I
needed to handle them differently for the white space counting.

I'll fix it.

PR #192, opened 2016-02-13
Add support for specify library color

andyscott:
This lets you specify library colors. If you don't provide a value, one will be automatically picked for you.

The logic for automatically picking colors is up for grabs. I wrote a mediocre solution to hold us over for the time being.


andyscott:
Force pushed a fix for a small mistake in the test code.

raulraja:
LGTM

PR #193, opened 2016-02-13
Use proper library for a given section when loading library index page

andyscott:
We were always using the first library when we needed to actually search for the correct one. This is a quick little fix for the bug.


raulraja:
LGTM

PR #194, opened 2016-02-13
Support evaluating exercise

andyscott:
This should ultimately bring support for evaluating simple exercises.

I've updated the compiler to output the fully qualified method name for evaluating the exercise. An example path might be `stdlib.ClassExercises.classExercise1` where `stdlib` is a package, `ClassExercises` is an object implementing the `exercise.Section` trait, and `classExercise1` is a method that can be invoked via reflection.

I am blocked on implementing the rest of this as I am unable to log in on a locally running instance and the eval route requires authentication. How should I proceed?


raulraja:
LGTM. @dialelo can you take care of the auth problem after I merge this in? Thanks!

Issue #168, opened 2016-02-10
Add license footer/header


FPerezP:
We need to add a license text to all our files. 

@raulraja, a couple of questions to answer before considering it ready to work on:
* What's the text? Do we have a template? What kind of license are we going to set? 
* Are we going to add it as a footer or as a header?

raulraja:
I'll work on it. The license will be ASL2

andyscott:
I have some code ready to add this automatically to all our files with the sbt header plugin. I've used roughly the same snippet for several projects.

PR #198, opened 2016-02-15
Add license; adjust headers for most sources files to include copyright

andyscott:
Getting the ball rolling on #168.

Because this change is big, if we want any adjustments I will just force push a new commit to keep everything cleanly in a single commit.


dialelo:
LGTM, better to have this merged as soon as possible so you don't have to update it all the time.

PR #199, opened 2016-02-15
UI fixes for code evaluation

dialelo:
- [x] A better style for the compile button which doesn't overflow with one-line exercises
- [x] Control compile button's opacity depending on compilation state
- [x] A fluid footer for the exercises to better reflect the original design 
- [x] Roll back to Postgres for development due to H2 Postgres emulation not working as expected

I'm rolling back to Postgres for development in this PR because it was causing exceptions when trying to insert in tables with relations. I've included a script in `scripts/create_dev_db` that creates the development DB.


rafaparadela:
Looks good to me

PR #206, opened 2016-02-16
Enable code evaluation

andyscott:
Almost ready. I'll ping you when it's time to review.


andyscott:
This is ready now.

@juanpedromoreno @rafaparadela or @raulraja please review

andyscott:
Bonus: You can evaluate expressions with this update. `1 + 1` will work for parameters of type `Int`, `!false` would work for `Boolean`, and `List(1, 2, 3, 4).tail` would work for `Seq[Int]`.

raulraja:
@andyscott this is amazing. Great work. LGTM!

PR #217, opened 2016-02-17
Don't accept runtime errors on evaluation as good

dialelo:


dialelo:
Please review and merge @rafaparadela, @andyscott, @raulraja or @juanpedromoreno.

juanpedromoreno:
Looks good :+1:

site/server/app/com/fortysevendeg/exercises/controllers/ExercisesController.scala
```diff
@@ -43,8 +43,11 @@ class ExercisesController(
               } yield exerciseEvaluation
 
               eval.runTask match {
-                case Xor.Right(result) ⇒ Ok("Evaluation succeeded : " + result)
-                case Xor.Left(error)   ⇒ BadRequest("Evaluation failed : " + error)
+                case Xor.Right(response) ⇒ response match {
```
raulraja: I'm not sure bad request is the right http status code for what we are doing but I guess it works for the time being though we should investigate what other http status fit better for failed evaluation. Also this doubly nested Xor could possibly be flattened here to avoid the nested match cases.

raulraja:
LGTM, though made a couple of comments that we should consider for future improvements.

PR #219, opened 2016-02-17
User progress in the UI

dialelo:
- [x] Fetch and display current library and sections' exercise solutions
- [ ] Display current library's progress meter accurately
- [ ] Display section's bullet point differently for completed and incompleted sections  
- [ ] Display modal if an unauthenticated user tries to evaluate an exercise

Not ready for merge yet but opening for receiving feedback.

### DIFF
--- a/site/client/src/main/scala/Api.scala
+++ b/site/client/src/main/scala/Api.scala
@@ -21,7 +21,15 @@ import org.scalajs.dom.ext.{ Ajax, AjaxException }
 object Client {
   def readProgress(library: String, section: String, raw: String): List[ClientExercise] = {
     val parsedBody = read[LibrarySectionArgs](raw)
-    parsedBody.exercises.map(e ⇒ ClientExercise(library, section, e.methodName, arguments = e.args))
+    parsedBody.exercises.map(e ⇒ {
+      ClientExercise(
+        library = library,
+        section = section,
+        method = e.methodName,
+        arguments = e.args,
+        state = if (e.succeeded) Solved else Unsolved
+      )
+    })
   }
 
   def fetchProgress(library: String, section: String): Future[Option[List[ClientExercise]]] = {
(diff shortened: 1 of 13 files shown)


site/client/src/main/scala/Api.scala
```diff
@@ -18,6 +19,29 @@ import common.ExtAjax._
...
+        method = e.methodName,
+        arguments = e.args,
+        state = if (e.succeeded) Solved else Unsolved
+      )
+    })
+  }
+
+  def fetchProgress(library: String, section: String): Future[Option[List[ClientExercise]]] = {
+    val url = Routes.Exercises.progress(library, section)
+    Ajax.get(url).map(r ⇒ {
```
raulraja: This is essentially map + filter which may be simplified via collect and partial function.

site/client/src/main/scala/Effects.scala
```diff
@@ -16,17 +18,29 @@ object Effects {
...
-    case _ ⇒ noop
+    case Start()                 ⇒ loadInitialData
+    case CompileExercise(method) ⇒ compileExercise(s, method)
+    case _                       ⇒ noop
+  }
+
+  def loadInitialData: Future[Option[Action]] = {
+    DomHandler.libraryAndSection.fold(Future(None): Future[Option[Action]])(libAndSection ⇒ {
+      val (lib, sect) = libAndSection
+      Client.fetchProgress(lib, sect).map(maybeState ⇒ {
```
raulraja: This may also be simplified via collect and partial function to avoid the doubly nested map

site/client/src/main/scala/Ui.scala
```diff
@@ -15,74 +20,86 @@ import model.Exercises._
 import utils.DomHandler._
 
 object UI {
+  val UNSOLVED = "unsolved"
```
raulraja: This may be also modeled with an ADT of case objects for better type safety down the road if these constants are reused.

raulraja: `case object Unsolved extends ExerciseState("unsolved"). Also there may be a use case for symbols here.

dialelo:
I'm reopening this as a separate PR to target the `version-2` branch. @raulraja thanks for the comments, i'll address them in the new PR.

PR #225, opened 2016-02-18
Show a user's exercise progress

dialelo:
- [x] Depends on the server sending the progress of every exercise, not just the evaluated ones

Needs an error message when exercise fails due to the user not being authenticated but I think we should leave that for the Github auth issue.

### DIFF
--- a/site/client/src/main/scala/Actions.scala
+++ b/site/client/src/main/scala/Actions.scala
@@ -8,7 +8,7 @@ package actions
 import model.Exercises._
 
 sealed trait Action
-case class NoOp() extends Action
+case class Start() extends Action
 case class SetState(s: State) extends Action
 case class UpdateExercise(method: String, args: Seq[String]) extends Action
 case class CompileExercise(method: String) extends Action
--- a/site/client/src/main/scala/Actions.scala
+++ b/site/client/src/main/scala/Actions.scala
@@ -8,7 +8,7 @@ package actions
 import model.Exercises._
 
 sealed trait Action
-case class Start() extends Action
+case object Start extends Action
 case class SetState(s: State) extends Action
 case class UpdateExercise(method: String, args: Seq[String]) extends Action
 case class CompileExercise(method: String) extends Action
--- a/site/client/src/main/scala/Effects.scala
+++ b/site/client/src/main/scala/Effects.scala
@@ -18,7 +18,7 @@ object Effects {
   def noop: Future[Option[Action]] = Future(None)
 
   def perform(s: State, a: Action): Future[Option[Action]] = a match {
-    case Start()                 ⇒ loadInitialData
+    case Start                   ⇒ loadInitialData
     case CompileExercise(method) ⇒ compileExercise(s, method)
     case _                       ⇒ noop
   }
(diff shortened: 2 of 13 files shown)


site/client/src/main/scala/Effects.scala
```diff
@@ -16,17 +18,29 @@ object Effects {
...
+    case Start()                 ⇒ loadInitialData
+    case CompileExercise(method) ⇒ compileExercise(s, method)
+    case _                       ⇒ noop
+  }
+
+  def loadInitialData: Future[Option[Action]] = {
+    DomHandler.libraryAndSection.fold(Future(None): Future[Option[Action]])(libAndSection ⇒ {
+      val (lib, sect) = libAndSection
+      Client.fetchProgress(lib, sect).collect({
+        case Some(state) ⇒ Some(SetState(state))
```
raulraja: We should try Avoiding explicit use of Some because `Option(null) == None` but `Some(null) == Some(null) `

raulraja: Not saying this is the case but it can eventually bite us.

site/client/src/main/scala/State.scala
```diff
@@ -10,6 +10,7 @@ import actions._
 
 object State {
   def update(s: State, a: Action): State = a match {
+    case Start()                      ⇒ List()
```
raulraja: Nil

site/client/src/main/scala/Effects.scala
dialelo: Will keep that in mind :+1:

site/client/src/main/scala/Actions.scala
```diff
@@ -8,7 +8,7 @@ package actions
 import model.Exercises._
 
 sealed trait Action
-case class NoOp() extends Action
+case class Start() extends Action
```
raulraja: If this has no properties can it just be a case object?

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -79,7 +79,8 @@ class ApplicationController(
   def javascriptRoutes = Action { implicit request ⇒
     Ok(
       JavaScriptReverseRouter("jsRoutes")(
-        routes.javascript.ExercisesController.evaluate
+        routes.javascript.ExercisesController.evaluate,
```
raulraja: `import routes.javascript._`

site/server/app/views/templates/library/index.scala.html
```diff
@@ -1,4 +1,4 @@
-@(library: shared.Library, section: shared.Section)(implicit request: RequestHeader)
+@(library: shared.Library, section: shared.Section, progress: Int = 0)(implicit request: RequestHeader)
```
raulraja: I recall there being a warning by the use of default args values in templates. We should look into it.

raulraja:
Minor comments. LGTM!  Great stuff!

rafaparadela:
I've been testing locally and it works smoothly. Great!

PR #226, opened 2016-02-18
Various Compiler Updates

andyscott:
- clean up code
- remove `Option` wrapper on various fields in the runtime model

### DIFF
--- a/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/Compiler.scala
+++ b/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/Compiler.scala
@@ -137,8 +137,8 @@ case class Compiler() {
                 treeGen.makeExercise(
                   name = exerciseInfo.comment.name,
                   description = exerciseInfo.comment.description,
-                  code = Some(exerciseInfo.code), // TODO: remove wrapper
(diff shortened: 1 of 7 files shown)


raulraja:
LGTM

andyscott:
I was going to add some more to this, but you guys were too quick!

juanpedromoreno:
I'm sorry @andyscott, I needed that code :/ . Please, work on a different branch. Thanks.

raulraja:
Also it looks like this breaks compilation of the std lib exercises

andyscott:
image

I believe you're picking up an old version of the compiler plugin, which has been published locally to your ivy2 repo.

There are a few ways to fix this. I would most certainly recommend doing a `clean` of the content project.

     > cd site; sbt ';project content; clean'

Now you need to either republish the compiler plugin or clear out your ivy2 repo. If you run the the scripted task in the core project it will automatically do this.

    > cd core; sbt scripted

or, if you want to just remove your ivy2 repo...

    > rm -r ~/.ivy2/local/com.47deg/
    > rm -r ~/.ivy2/cache/com.47deg/

andyscott:
I'll add that I don't know why this is happening now. It used to happen to me when I was originally setting up the `site` and `core` projects. Once I get the time to do some official cleanup of the structure, it should be less of an issue.

raulraja:
We should add that to our README

PR #232, opened 2016-02-19
Simplifies the exercise evaluation result

juanpedromoreno:
This PR is a pair programming between @raulraja and me.

Please, @andyscott @dialelo review when you have a chance. Thanks.

### DIFF
--- a/site/server/app/com/fortysevendeg/exercises/controllers/ExercisesController.scala
+++ b/site/server/app/com/fortysevendeg/exercises/controllers/ExercisesController.scala
@@ -5,10 +5,10 @@
 
 package com.fortysevendeg.exercises.controllers
 
-import cats.data.Xor
 import com.fortysevendeg.exercises.app._
 import com.fortysevendeg.exercises.persistence.domain.SaveUserProgress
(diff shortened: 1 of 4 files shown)


andyscott:
@raulraja I think the ADT is the way to go. As it stands, this PR undoes some of the work yesterday that addressed specifically capturing the exceptions thrown in the eval'd method. For example, if the exercise is:
```scala
true should be(__)
```
and you enter `false`, then code will compile and run but you will still get an error that indicates otherwise.

raulraja:
Feel free to discard the PR but we need the simplify the nested Xor now
because a bug was discovered when reporting it from the lack of proper
pattern matching on the result. I don't think it's OK to expect the call
site for this method to provide exhaustive matching for double nested Xor
Ior paths. If there is a way to express all the errors and still obtain a
single List[Ex] Xor A then that is what we want.

andyscott:
The nesting definitely should be simplified.

andyscott:
See preliminary work in #236.

juanpedromoreno:
This PR is covered by https://github.com/47deg/scala-exercises/pull/236.

PR #233, opened 2016-02-19
Allow user emails to be null

andyscott:
Users may set their GitHub email to private, which results in a `null` in the JSON payload received during the Oauth cycle. (See https://developer.github.com/v3/users/#get-a-single-user)

Previously an exception was thrown because the unmarshaller didn't expect the null. This fix changes the email field to an `Option[String]`, after demarshalling.

It also lets me log in to the app, because I'm one of said users!


dialelo:
Good catch! Looks good to me so feel free to merge.

Issue #229, opened 2016-02-19
Fix authentication issues


rafaparadela:
There are some points related to authentication that could be improved:

- [x] Sometime, although the button to login seems available to start the process, it doesn't work because the user is already logged. For that reason, visiting the `/logout` endpoint usually solves the inconsitent status. However, the button should show the real auth status accordingly.
- [x] On the other hand, in Exercises screen, the auth button should show the actual status as well, instead of the mocked partial template with my name/face.
- [ ]  Due to exercises shouldn't be evaluated for unlogged users, we should implement this logic previously to the `evaluation` call. And show a modal with the invatation for signing up.

PR #234, opened 2016-02-19
Fix authentication issues

rafaparadela:
This PR resolves issues exposed in #229:

- [x] Sometime, although the button to login seems available to start the process, it doesn't work because the user is already logged. For that reason, visiting the /logout endpoint usually solves the inconsistent status. However, the button should show the real auth status accordingly.
- [x] On the other hand, in Exercises screen, the auth button should show the actual status as well, instead of the mocked partial template with my name/face.
- [ ] Due to exercises shouldn't be evaluated for unlogged users, we should check this condition previous to the evaluation call. And show a modal with the invitation for signing up.

### DIFF
--- a/site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
+++ b/site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
@@ -31,25 +31,17 @@ class ApplicationController(
 
   def index = Action.async { implicit request ⇒
 
-    val scope = "user"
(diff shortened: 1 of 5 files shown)


rafaparadela:
@raulraja @dialelo Could you take a look please? Thanks.

dialelo:
The changes look good to me! However, the build failed [because an ambiguous jquery version](https://travis-ci.org/47deg/scala-exercises/jobs/110541101#L3396).

raulraja:
LGTM

PR #236, opened 2016-02-19
Convert eval result to ADT

andyscott:
This initial commit will be broken (in the site project). I will fix it shortly.

### DIFF
--- a/core/runtime/src/main/scala/com/fortysevendeg/exercises/MethodEval.scala
+++ b/core/runtime/src/main/scala/com/fortysevendeg/exercises/MethodEval.scala
@@ -15,28 +15,77 @@ import cats.std.list._
 import cats.syntax.flatMap._
 import cats.syntax.traverse._
 
+object MethodEval {
+
+  /** An evaluation result, which can have three possible results. */
+  sealed abstract class EvaluationResult[A](val didRun: Boolean) extends Product with Serializable {
+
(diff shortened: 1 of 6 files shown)


core/runtime/src/main/scala/com/fortysevendeg/exercises/MethodEval.scala
```diff
@@ -15,28 +15,43 @@ import cats.std.list._
 import cats.syntax.flatMap._
 import cats.syntax.traverse._
 
+object MethodEval {
+  sealed abstract class EvaluationResult extends Product with Serializable {
+    def didRun: Boolean
```
raulraja: What about a constructor arg?

raulraja:
So far LGTM!

andyscott:
- I'd like to get rid of the nested `Xor`s in the site projects. This requires that the ADT in the runtime project is made available in that scope or we create a parallel ADT for the site projects. I wasn't sure what route we wanted to go, so I left the Xors.
- This should fix the issue in #232. Now, the right projection of the eval result indicates a full success. The original logic for updating user progress should work correctly, as it originally assumed this about the Xor.

raulraja:
@juanpedromoreno can you review and merge this?

juanpedromoreno:
Looks good!

PR #238, opened 2016-02-21
JP SE-224 Unifies the persistence layer for User and User Progress domains

juanpedromoreno:
Please, @raulraja @dialelo Review when you have a chance. Thanks.

PR #239, opened 2016-02-21
Changes name field as option

rafaparadela:
According to the possibility that an user may not have any defined name, the app should cover this case. This PRs resolves this issue.

@juanpedromoreno @raulraja Is good to merge?

### DIFF
--- a/site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
+++ b/site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
@@ -82,7 +82,7 @@ class OAuth2Controller(
         withHeaders(HeaderNames.AUTHORIZATION → s"token $authToken").
         get().map { response ⇒
           val login = (response.json \ "login").as[String]
-          val name = (response.json \ "name").as[String]
+          val name = (response.json \ "name").asOpt[String]
           val githubId = (response.json \ "id").as[Long]
           val avatarUrl = (response.json \ "avatar_url").as[String]
           val htmlUrl = (response.json \ "html_url").as[String]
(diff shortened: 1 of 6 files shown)


site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
```diff
@@ -82,7 +82,7 @@ class OAuth2Controller(
         withHeaders(HeaderNames.AUTHORIZATION → s"token $authToken").
         get().map { response ⇒
           val login = (response.json \ "login").as[String]
```
raulraja: This is a really bad way to parse de json it should all be marshaled and validated at once, as is it may throw an exception if the incoming messages get malformed due to API changes on the Github side.

raulraja: There are several examples in the play docs with example on how to properly write a custom writer or reader

site/server/test/ArbitraryInstances.scala
```diff
@@ -33,7 +33,7 @@ trait ArbitraryInstances extends Assertions {
       email ← Gen.alphaStr
     } yield Request(
       login = login.toString,
-      name = name,
+      name = Some(name),
```
raulraja: Please prefer the option type instead of some because it's different behavior on null values

raulraja:
LGTM and most of the comments are related to the old POC code on the Github Integration which was pretty bad not anything you've done but I would like this addressed in this PR or another one with new issues associated.

raulraja:
Let's address it in a separate PR so we can get also #238 in

rafaparadela:
I agree with you, indeed there is a ticket to work on it: https://github.com/47deg/scala-exercises/issues/113

PR #242, opened 2016-02-21
JP - Unifies the arbitrary instances in the PersistenceModuleSpec file

juanpedromoreno:
This PR accomplish the technical debt acquired the last week. Now, all our persistence tests are using the same arbitrary instances together with the shapeless scalacheck impl. Please, @raulraja could you review? Thanks.


raulraja:
LGTM

Issue #240, opened 2016-02-21
Missing Compile Button


juanpedromoreno:
The `Compile` button doesn't appear in this case:

http://jpedrom.es/1SJqEu9

rafaparadela:
@juanpedromoreno In this particular cases shown in the screen captured, the "compile" button is rightly hidden:
- It's hidden in the first exercise because it's successfully resolved (if you change it, then should appear the button).
- In the second one, because not all the inputs are filled.

Anyway, I will take a look to search any unexpected behaviour.

juanpedromoreno:
Thanks @rafaparadela . It makes sense.

rafaparadela:
@juanpedromoreno After researching, certainly an unexpected behaviour occurs that is explained and resolved in this PR #243

PR #243, opened 2016-02-21
Set the exercise as Unsolved when it's changed after a succeeded evaluation

rafaparadela:
As a result of the issue #240, I noticed that once an exercises is successfully evaluated, is not possible running the evaluation again.
I guess the right behaviour is allow that exercises can be evaluated once and again. So this PR set the exercise as unsolved when is changed (although it had been set as solved previously), allowing to run the evaluation again.

@raulraja @dialelo Does this make sense for you?


raulraja:
LGTM

PR #257, opened 2016-02-22
JP - SE-222 Persistence args as Array

juanpedromoreno:
Please, @raulraja @dialelo Review when you have a chance. Thanks!

Ticket https://github.com/47deg/scala-exercises/issues/222

BTW, @dialelo I've opened a separate issue https://github.com/47deg/scala-exercises/issues/256 to tackle what we talked this morning about the case object enums.


site/server/app/com/fortysevendeg/exercises/persistence/domain/UserProgressQueries.scala
```diff
@@ -24,7 +24,7 @@ object SaveUserProgress {
       method:       String,
       version:      Int,
       exerciseType: ExerciseType,
-      args:         Option[String],
+      args:         Option[List[String]],
```
raulraja: Can't a postgress nullable value map to an empty List instead of double nesting type constructors with empty capable semantics?

site/shared/src/main/scala/shared/User.scala
```diff
@@ -19,8 +19,8 @@ case class UserProgress(
   sectionName:  String,
   method:       String,
   version:      Int,
-  exerciseType: String         = "Other",
-  args:         Option[String],
+  exerciseType: String               = "Other",
+  args:         Option[List[String]],
```
raulraja: I'm not convinced that this double nesting is required and I think it can be considered technical debt with an appropriate issue logged if this isn't solved in this PR.

raulraja:
LGTM once my comments are addressed

dialelo:
Same observation as @raulraja, and we may want to give the "args" field a default value. Other than that LGTM!

juanpedromoreno:
I see your point, @raulraja @dialelo . I tried it, however, take a look to this `Lamentations of NULL` section https://tpolecat.github.io/doobie-0.2.0/09-Arrays.html . I'm not sure if we could flatten this double nesting in somehow. Any thoughts?

juanpedromoreno:
I've just submitted a new commit making the `args` field not nullable to avoid the double nesting. @raulraja @dialelo does it ready to merge?

dialelo:
Excellent, LGTM :shipit:

PR #261, opened 2016-02-23
Configure top library order for homepage

dialelo:
Added a new configuration option: `exercises.top_libraries`, that will contain an array with the names of the libraries we want to show on top in the homepage. @raulraja @juanpedromoreno please take look!

### DIFF
--- a/site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
+++ b/site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
@@ -6,14 +6,17 @@
 package com.fortysevendeg.exercises.controllers
 
 import java.util.UUID
+import scala.collection.JavaConverters._
 
 import cats.data.Xor
 import com.fortysevendeg.exercises.app._
 import com.fortysevendeg.exercises.services.free._
+import com.fortysevendeg.exercises.services.ExercisesService
 import com.fortysevendeg.exercises.services.interpreters.ProdInterpreters
 import com.fortysevendeg.exercises.utils.OAuth2
 import com.fortysevendeg.shared.free.ExerciseOps
 import doobie.imports._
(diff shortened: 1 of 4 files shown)


site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
```diff
@@ -30,12 +33,15 @@ class ApplicationController(
     userProgressOps: UserProgressOps[ExercisesApp],
     T:               Transactor[Task]
 ) extends Controller with AuthenticationModule with ProdInterpreters {
+  implicit def application: Application = Play.current
+
+  lazy val topLibraries: List[String] = application.configuration.getStringList("exercises.top_libraries") map (_.asScala.toList) getOrElse Nil
```
juanpedromoreno: `application.configuration.getStringList("exercises.top_libraries").asScala` would work as well.

juanpedromoreno:
Minor comment. LGTM.

site/server/app/com/fortysevendeg/exercises/controllers/ApplicationController.scala
dialelo: Great suggestion, going to change it right away.

dialelo: Oops, didn't work. Take into account that `getStringList` returns an option.

juanpedromoreno: Uou! So the Play config seems to be different to the typesafe config library :/

https://github.com/typesafehub/config/blob/ba5e0da6941fd67b779d97c3bd82f1a9dfcfb657/config/src/test/scala/com/typesafe/config/impl/ConfigTest.scala#L584

Hence LGTM!

Issue #262, opened 2016-02-23
Fix UI issues previous code-freeze


rafaparadela:
We should solve these issues:

- [x] Reformat code blocks with a properly theme.
- [x] Rename Libraries and Section, replacing camelCase by lower_case/underscore
- [x] Set placeholders for "Contributors" & "Commit" sections.
- [x] Replacing `res` by inputs elements only within exercises code
- [ ] ~~Add Section name at `breadcrumbs`.~~
- [x] Enable all the exercises of StdLib, with their type signatures properly defined

PR #263, opened 2016-02-23
Fix UI issues previous code freeze

rafaparadela:
This PR solves #262 

- [x] Reformat code blocks with a properly theme.

- [x] Rename Libraries and Section, replacing camelCase by lower_case/underscore

- [x] Set placeholders for "Contributors" & "Commit" sections.

- [x] Replacing `res` by inputs elements only within exercises code
- [x] ~~Add Section name at `breadcrumbs`.~~
- [x] Enable all the exercises of StdLib, with their type signatures properly defined

@raulraja @dialelo @andyscott @juanpedromoreno Could you review please?


core/compiler/src/test/scala/com/fortysevendeg/exercises/compiler/DocParserSpec.scala
```diff
@@ -1,12 +1,11 @@
 package com.fortysevendeg.exercises
 package compiler
 
-import cats.data.Xor
-
+//import cats.data.Xor
 import org.scalatest._
```
raulraja: I'f these are no longer needed we should get rid of them

raulraja:
LGTM and a million thanks for migrating all the stdlib exercises.

PR #264, opened 2016-02-23
Format code samples with Scalariform

dialelo:
- Had to wrap the code in an `object` because the code to format has to be in an object or a class and not all examples are like that. When unwrapping I make sure to trim the leading whitespace that Scalariform adds to the body of the object.

- I run the formatting in`Xor.catchNonFatal` because the `format` method happily throws exceptions. When having a piece of non-Scala code (for example a session in the REPL) or code that can't be parsed we return it verbatim.

@raulraja @rafaparadela @juanpedromoreno @andyscott take a look please!

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -24,7 +24,9 @@ lazy val compiler = (project in file("compiler"))
     exportJars      := true
   )
   .settings(commonSettings: _*)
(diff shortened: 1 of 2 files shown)


raulraja:
LGTM!

PR #265, opened 2016-02-23
Show the completed sections instead of the completed exercises

dialelo:
@juanpedromoreno @raulraja take a look, the problem with the `findByUserIdAggregated` is that it retrieves the completed exercise count and what we want is the count of completed sections.


juanpedromoreno:
LGTM!!

Issue #260, opened 2016-02-23
Don't spoil users with the right exercise solutions when they don't guess it 


dialelo:
When a method evaluation fails due to an assertion error, we are reporting the message returned by the matcher that fails. This means that when failing the matchers will show a helpful diff with the expectations vs reality, see the following caption:

![f](https://cloud.githubusercontent.com/assets/409039/13250819/8bb0b890-da2c-11e5-9389-6d69ebdf45c5.png)

This messages are helpful in tests but we don't want to spoil users with the right answers when they fail.

We want to provide better error messages that don't include the correct results. Ideally this should be done in the code that runs the exercises.

PR #266, opened 2016-02-23
Change bubble up behavior of scalatest exceptions in exercises

andyscott:
This updates the core project so that exercise exceptions thrown by scalatest are handled in a special manner.

I chose to change the exercise evaluation result to a new type. The left projection is a simple error message, while the right projection is the success value. This simplifies logic in the exercise controller.


dialelo:
LGTM and works perfectly, closes #260.

andyscott:
Added note: this solution will run into issues when we eventually move exercises into separate classloaders. The reason being that scalatest will be specific to the exercise classpath. It would be unavailable or potentially of a conflicting version if used from the server classpath (as it is in this code).

PR #274, opened 2016-02-24
Compile exercise on blur

dialelo:
- Disallows tab navigation to the "Run" buttons.
- Tries to compile exercises on blur as it used to

@rafaparadela @israelperezglez please take a look!


rafaparadela:
LGTM!

PR #275, opened 2016-02-24
Lib desc updates

raulraja:
@rafaparadela @dialelo Can you please review?


dialelo:
LGTM, I've pushed a commit with minor wording changes.

By the way, we should limit the size of the boxes when the descriptions are too large. Also, longer descriptions make library boxes bigger @israelperezglez @rafaparadela:

descs

dialelo:
Just pushed a CSS fix I wrote with @israelperezglez  for making all the boxes have the same height.

dialelo:
Don't mind the [failure](https://travis-ci.org/47deg/scala-exercises/jobs/111429852) in the previous tests run, the build timed out waiting for `sbt` to download.

PR #282, opened 2016-02-25
Add a section for learning about functors to the Cats exercises

dialelo:
@raulraja please take a look.


raulraja:
This is awesome! LGTM!

PR #283, opened 2016-02-25
Rebased version 2

dialelo:
This is a rebase of `version-2` on top of `master` since it doesn't merge cleanly. I had a lot of conflict resolution to do and didn't get everything right so I [undid my mistakes with a new commit](https://github.com/47deg/scala-exercises/commit/88bfa103a8b6626289ef518ce955977f5d2672d4). This was basically checking out the files I accidentaly deleted or modified wrongly from `version-2`, if you think isn't acceptable i can rebase interactively and fix my mistakes in the right commits but it'll take some more time.

Since GitHub won't show a diff with master I recommend diffing this rebased version to the original `version-2`, you'll see that it only adds headers to a couple of files in the Shapeless exercise:

```sh
git fetch --all
git diff origin/version-2..origin/rebased-version-2
```


dialelo:
@raulraja @rafaparadela @juanpedromoreno @andyscott take a look please!

raulraja:
Ship it!

dialelo:
Updated with the changes from #284

dialelo:
:shipit:

andyscott:
There's an alternative to this-- not sure if it's considered sane or not--

We could just rename master to `version-1` or some other name, then rename `version-2` to `master`. It avoids a ridiculous rebase.

andyscott:
I don't think a rebase is the right move for this merge. Instead, we should merge master into `version-2` with the `ours` strategy.

> git checkout version-2
> git checkout -B version-2-megamerge
> git merge -s ours master

I've done those steps, and submitted a separate PR (#285) which you can look at. Next, we would simply merge the PR branch into master and remove any artifacts from the original master branch.

dialelo:
Closed in favour of #285

PR #285, opened 2016-02-25
Replace Master with Version 2

andyscott:
Steps performed:

> git checkout version-2
> git checkout -B as-v2-replace-master
> git merge -s ours master


dialelo:
:shipit:

PR #286, opened 2016-02-26
Use ExerciseType instead of Strings for type safety

### LINKED ISSUE #256: Technical Debt - Postgres Enums
We should use Meta instances to use the `ExerciseType` (UserProgress Table) case object with implicit auto-conversions to the varchar and vice versa.

dialelo:
- Adds a `Meta` instance so we can insert and read ExerciseType from the
  database
- Implements a reader for the JSON "exerciseType" attribute
- Modifies types coercing ExerciseType to/from String to use
  ExerciseType

Closes #256


dialelo:
@raulraja @juanpedromoreno @rafaparadela please take a look!

juanpedromoreno:
Outstanding work! LGTM.

dialelo:
:shipit:

juanpedromoreno:
:shipit:

PR #287, opened 2016-02-26
Support Explanations in Comments

andyscott:
This brings support for explanations in comments. Additionally, I've rewritten and cleaned up a lot of of the comment parsing/rendering code.

At the moment it's not done. I am going to squash together some of my commits before we do the final code review.


rafaparadela:
👏🏼

andyscott:
This will change things a bit. Libraries and Sections comments will no longer require a single line at the top with the name.

    /** This is the name
      *
      * This is the description
      */

_becomes_

    /** This is the description
      *
      * This is still more description
      * @param name This is the name
      */

Explanations might look like this:

    /** Description description description.
      * Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
      * tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
      * veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex.
      *
      * @param explanation in voluptate velit esse cillum dolore eu fugiat nulla
      * pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa
      * qui officia deserunt mollit anim id est laborum.
      * 
      * === still in the explanation tag ===
      * etc etc etc
      * 
      * @param name HelloWorld
      */

andyscott:
@rafaparadela @raulraja @dialelo this is ready for a review. In particular, please review the documentation in [comments.scala](https://github.com/47deg/scala-exercises/blob/as-support-explanations/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala) to make sure I described everything appropriately.

Thanks!

core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala
```diff
@@ -0,0 +1,341 @@
...
+
+  object Mode {
+    type Aux[N0[_], D0[_], E0[_]] = Mode {
+      type N[A] = N0[A]
+      type D[A] = D0[A]
+      type E[A] = E0[A]
+    }
+
+    /** Library comments. Required name. Required description.
+      * Require no description.
```
dialelo: I think you mean "explanation" instead of "description", could be? Also in the next type's comment.

core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala
```diff
@@ -0,0 +1,341 @@
...
+      import global._
+      override def parse(comment: DocComment) = {
+        val nowarnings = settings.nowarn.value
+        settings.nowarn.value = true
+        try parseAtSymbol(comment.raw, comment.raw, comment.pos)
+        finally settings.nowarn.value = nowarnings
+      }
+      override def parse(comment: String) = parse(DocComment(comment))
+      override def internalLink(sym: Symbol, site: Symbol): Option[LinkTo] = None
+      override def chooseLink(links: List[LinkTo]): LinkTo = links.headOption.orNull
```
dialelo: Not sure of the implications of the possibility of a `null` flowing around, wouldn't `Option` be safer?

dialelo:
Excellent work, LGTM once comments are addressed.

core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala
andyscott: You're right!

core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala
andyscott: It's based off code from scaladoc internals, so I just left it the way they had it.

How would you use an option here? I'm overriding a method where I have to return a `LinkTo`.

dialelo: :+1:

dialelo:
:shipit:

PR #288, opened 2016-03-01
Fix regressions introduced by #286

dialelo:
I implemented the JSON read wrongly, which broke the evaluation from client HTTP requests :blush: It's now fixed along with a minor fix to one of the exercises.


rafaparadela:
Nice. Thank you. LGTM

Issue #170, opened 2016-02-10
Show contributors in each exercises category


FPerezP:
We need to use Github API to get all contributors for each category so we can show them properly on each category page.

PR #290, opened 2016-03-03
Cleanup Oauth code

dialelo:
- [x] Implemented OAuth code in terms of user algebra instead of repository, moving the `getOrCreate` function to the `UserOps` primitives.
- [x] Cleaned up code in the OAuth controller, separating GitHub communication in their own methods. In the future we may want to split this further into a propper GH client, when we use GitHub data for showing contributors.
- [x] Improved the JSON parsing, it will no longer fail at runtime.
- [x] Changed OAuth scope so we have only access to data that is public

@raulraja @rafaparadela @juanpedromoreno take a look and let me know if something is missing.

### DIFF
--- a/site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
+++ b/site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
@@ -84,11 +84,11 @@ class OAuth2Controller(
   def githubUserRequest(authToken: String) =
     ws.url("https://api.github.com/user").withHeaders(HeaderNames.AUTHORIZATION → s"token $authToken").get()
 
-  def fetchGitHubUser(authToken: String): Future[Option[GitHubUser]] = for {
-    response ← githubUserRequest(authToken)
-  } yield response.json.validate[GitHubUser] match {
-    case ok: JsSuccess[GitHubUser] ⇒ Some(ok.get)
-    case _                         ⇒ None
+  def fetchGitHubUser(authToken: String): Future[Option[GitHubUser]] = {
+    githubUserRequest(authToken).map(_.json.validate[GitHubUser] match {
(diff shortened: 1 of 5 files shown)


site/server/app/com/fortysevendeg/exercises/utils/OAuth2.scala
```diff
@@ -34,79 +40,108 @@ object OAuth2 {
...
+    (JsPath \ "avatar_url").read[String] and
+    (JsPath \ "html_url").read[String] and
+    (JsPath \ "email").readNullable[String]
+  )(GitHubUser.apply _)
+
+  def githubUserRequest(authToken: String) =
+    ws.url("https://api.github.com/user").withHeaders(HeaderNames.AUTHORIZATION → s"token $authToken").get()
+
+  def fetchGitHubUser(authToken: String): Future[Option[GitHubUser]] = for {
+    response ← githubUserRequest(authToken)
```
juanpedromoreno: Maybe we could change this for comprehension for a single `map`.

juanpedromoreno:
Great work! Minor comment added and LGTM! 

I was wondering if it makes sense to move the `Github` stuff to a different algebra in  a different PR (oauth, contributors, etc. services in an isolated algebra).

rafaparadela:
Probably @juanpedro is right, but we're are thinking on making our own
github client isolated in other repo (there isn't any maintained github
client in Scala). So all these algebras that Juan Pedro mentioned would be
so simple.

rafaparadela:
Btw: looks good to me

dialelo:
> I was wondering if it makes sense to move the Github stuff to a different algebra in a different PR (oauth, contributors, etc. services in an isolated algebra).

Maybe, as @rafaparadela suggests, we can address this as part of #170

juanpedromoreno:
Sure :+1:

PR #291, opened 2016-03-04
Remove repeated articles from descriptions

murphyslaw:


dialelo:
Good catch, thanks for your time!

Issue #293, opened 2016-03-10
Avoid `java.lang.NoClassDefFoundError: scalariform/formatter/preferences/SpacesAroundMultiImports`


dialelo:
As @jackcviers pointed out in the Gitter channel, when downloading and trying to build the project a `java.lang.NoClassDefFoundError: scalariform/formatter/preferences/SpacesAroundMultiImports` error is thrown. This can be worked around adding the following incantation to the sbt plugin configuration:

```scala
dependencyOverrides in ThisBuild += "org.scalariform" %% "scalariform" % "0.1.8"
```

Doesn't this override the Scalariform dependency for every project? This solution is a workaround and we should properly fix this issue, not requiring everyone to fiddle with their sbt configuration. Downloading and building the project should just work.

jackcviers:
> dependencyOverrides in ThisBuild += "org.scalariform" %% "scalariform" % "0.1.8"

I put this in `/core/project/plugins.sbt` and `/site/project/plugins.sbt` and it resulted in the same error. What am I missing @dialelo?

dialelo:
I think you have to put it directly in your SBT configuration, I have it in `~/.sbt/0.13/plugins/plugins.sbt`. I think the project that's causing the conflict is `compiler` but I'm not quite sure, new to sbt/scala myself. /cc @andyscott

andyscott:
If you change your global plugins, I've also found it useful to clear out any cached files from SBT's global installation.

    cd ~/.sbt; \
        rm -r boot/; \
        find **/target -maxdepth 0 | xargs rm -r; \
        find **/staging | xargs rm -rf;

jackcviers:
Gotcha.

jackcviers:
Workaround
=========

As @andyscott and @dialelo  suggested, the issue seems to stem from having [ensime-sbt-cmd](https://github.com/ensime/ensime-sbt) in the global `~/.sbt/0.13/plugins.sbt`. Here's the full workaround I did to have my exercises and ensime too:

In `~/.sbt/0.13/plugins.sbt`, if the following line exists, comment it out:

```scala
addSbtPlugin("org.ensime" % "ensime-sbt" % "0.4.0")
```

and add the [fix](https://github.com/47deg/scala-exercises/issues/293#issue-139826258) from @dialelo above:

```scala
dependencyOverrides in ThisBuild += "org.scalariform" %% "scalariform" % "0.1.8"
```

and save the file.

[...]

PR #300, opened 2016-03-16
JP SE-293 Transitive Dependency issues with Ensime/Scalariform

juanpedromoreno:
This PR fixes, on one hand, the conflicting cross-version suffixes in org.scalariform (2.10, 2.11). On the other hand, try to clarify what could the user do in case of runtime exceptions.

Fixes: https://github.com/scala-exercises/scala-exercises/issues/293

Please, @raulraja @dialelo Could you review? Thanks.


andyscott:
Looks good to me. LGTM

Issue #259, opened 2016-02-23
Add a mechanism for converting user input to custom types


dialelo:
For writing our own exercises and specially for third party libraries' exercises we need a way to specify how to convert an input string from the user to an arbitrary type. Built-in types like `String` or `Int` are supported but we often have exercises like this:

```scala
def anExercise(res0: Xor.Right[String, Int]) = {
  val right: String Xor Int = Xor.Right(42)
  right should be(res0)
}
```

We can't currently convert an user-provided input to a `cats.data.Xor` and we should be able to do so. Same for the every type constructors that cats or other libraries provide.

Another case that we don't handle currently is referring to an identifier, see the following example:

```scala
def anExercise(res0: ?) = {
  val one = 1
  val anotherOne = 1
  (one == res0) should be(true)
}
```

Here, we expect the user to insert `anotherOne`. The problem is that we don't want to treat `res0`'s content as a string but an identifier that will be resolved to the value it refers to at runtime.

@andyscott feel free to split this into separate issues if you think they should be tackled separately.

@raulraja @rafaparadela @juanpedromoreno feel free to add your comments.

andyscott:
Both of these items are addressable. The difference, from what I can tell is that in one case _**(A)**_ you want to leverage types and symbols imported into the file and potentially into the object containing the exercises. In the other, **_(B)_** you want the user to have access to symbols that are local to the exercise's method body.

I strongly suspect I can tackle _**A**_ before the conference. The most obvious solution for this is to capture all imports at all static _(read: package, object, anything that does not get instantiated... so no classes or traits... I made this term up)_ scopes up until the exercise method. When the user input is parsed to a tree and typed, I just just ensure all those imports are present first.

I intend to eventually support **_B_**, but this will require dynamically generating and evaluating a Scala AST with the user's input inserted into the tree for the method. This solution will also work for A; so any work I do to quickly support just A will ultimately get thrown away when I do full support for B.

It's possible, if I push it, that I could get B done in time for the conference. But I'm not sure.

Thoughts? Which route should I take?

dialelo:
> I strongly suspect I can tackle A before the conference. The most obvious solution for this is to capture all imports at all static (read: package, object, anything that does not get instantiated... so no classes or traits... I made this term up) scopes up until the exercise method. When the user input is parsed to a tree and typed, I just just ensure all those imports are present first.

[...]

PR #304, opened 2016-03-18
Capture imports for koan style exercises

andyscott:
WIP for #259.

This has pushed the compiler code to be a bit messier than I want. Cleanup will come sometime soon, hopefully.


andyscott:
@raulraja I **think** this works for the desired cases.

I'll stress again that I'm not really fond of some the changes I had to make to get this to work. Cleanup to come down the road. Preferably after we publish artifacts and speed up the build and test process.

raulraja:
A much needed fix, thanks! LGTM!

LGTM

PR #308, opened 2016-03-21
Refactor user progress

dialelo:
- [x] Add a new primitive to the User progress ADT: `GetExerciseEvaluations`. Given a user, library and section gives you a list with the evaluations of the user.
- [x] Implement operations that depend on progress and exercises (only available at runtime) using Free instead of calling repositories explicitly from `userprogressop.scala`
- [x] Add a new constraint to user progress so we don't save duplicated exercise evaluations
- [x] Make sure tests run and pass

Still have to fix an issue with the repository tests but is ready for review, please take a look @raulraja.


raulraja:
Impressive refactor! LGTM

dialelo:
The tests now run and pass in CI. @scala-exercises/core-mergers please review and :shipit:

site/server/app/com/fortysevendeg/exercises/services/free/userprogressops.scala
```diff
@@ -28,137 +34,132 @@ final case class UpdateUserProgress(
...
-                sections = completedSections,
-                totalSections = totalSections,
-                completed = succeeded && completedSections == totalSections
-              )
-            }
-      }
-      list ← Free.freeMonad[F].sequence(items)
-    } yield OverallUserProgress(list)
+      allLibraries ← EO.getLibraries
+      libraryProgress ← Free.freeMonad[F].sequence(allLibraries.map(l ⇒ {
```
raulraja: Instead of explicitly using the `Free` monad instance here should we require this via implicit resolution in case there is a user provided instance for `Monad[Free[F, ?]]`?
This would be a good time to introduce Kind Projector too.

raulraja:
LGTM!

dialelo:
Added kind-projector and required `Free` instance through an implicit, will merge if you think it's fine @raulraja.

raulraja:
LGTM!

PR #310, opened 2016-03-21
Show helpful error message when there are network problems

### LINKED ISSUE #306: Show a helpful error message when the server doesn't respond to AJAX calls
When the server is down or a network requests times out, the client assumes it was an evaluation error. We should detect the cases when evaluation fails due to network conditions and show an error message like the following:

> We couldn't evaluate the exercise, try again in a few minutes.

dialelo:
Closes #306 

@raulraja take a look please.

### DIFF
--- a/site/client/src/main/scala/Api.scala
+++ b/site/client/src/main/scala/Api.scala
@@ -48,6 +48,17 @@ object Client {
         EvaluationResult(true, e.method)
       else
         EvaluationResult(false, e.method, r.responseText)
-    }).recover({ case exc: AjaxException ⇒ EvaluationResult(false, e.method, exc.xhr.responseText) })
+    }).recover({
+      case exc: AjaxException ⇒ {
+        EvaluationResult(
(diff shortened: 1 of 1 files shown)


site/client/src/main/scala/Api.scala
```diff
@@ -48,6 +48,17 @@ object Client {
...
       else
         EvaluationResult(false, e.method, r.responseText)
-    }).recover({ case exc: AjaxException ⇒ EvaluationResult(false, e.method, exc.xhr.responseText) })
+    }).recover({
+      case exc: AjaxException ⇒ {
+        EvaluationResult(
+          false,
+          e.method,
+          if (exc.isTimeout)
+            "We couldn't evaluate the exercise, please try again in a few minutes."
```
raulraja: `We couldn't evaluate your exercise. You may be experience internet connectivity issues`

raulraja:
LGTM

Issue #326, opened 2016-03-30
Evaluation fails with `java.lang.NoClassDefFoundError`


dialelo:
When I try to evaluate exercises I'm consistently getting a `java.lang.NoClassDefFoundError` exception.

![eval](https://cloud.githubusercontent.com/assets/409039/14143968/2eb759e0-f68e-11e5-9c93-4c8a84d66310.png)

@andyscott it might have been introduced by the recent changes in the exercise code generation?

PR #327, opened 2016-03-31
Fix runtime/eval/compiler classpaths

andyscott:
This addresses #326. There are still other issues with evaluation.

Part of this is some cleanup/simplification/validation.


andyscott:
@scala-exercises/core-mergers take a look

dialelo:
Fixes the problem, great work!

PR #328, opened 2016-03-31
Switch to Monix Observables in the client

dialelo:
This PR implements the client UI's reactive cycle in terms of Monix Observables instead of Rx's. Monix observables have plenty of well-documented operators and it integrates with cats nicely, unlike Rx. It even has its own `Task` type which we may want to use in the server at some point.

I think the data flow is more clear now. If you disagree we can close this pull request.

@scala-exercises/core-mergers take a look please and let me know your thoughts.


raulraja:
LGTM!

PR #331, opened 2016-03-31
Fix behavior of code formatter for comments.

### LINKED ISSUE #309: Some sample code snippets are not extracted correctly
We have noticed that there are some code snippets that are not correctly extracted from the comments. See this example in the "Tuples" section:

blank

The code from the exercise looks like the following, i can't see anything wrong with it:

```scala
object Tuples extends FlatSpec with Matchers with exercise.Section {
  /** Scala tuple combines a fixed number of items together so that they can be passed around as a whole. They are one indexed. Unlike an array or list, a tuple can hold objects with different types but they are also immutable. Here is an example of a tuple holding an integer, a string, and the console:
    *
    * {{{
    * val t = (1, "hello", Console)
    * }}}
    *
    * Which is syntactic sugar (short cut) for the following:
    *
    * {{{
    * val t = new Tuple3(1, "hello", Console)
    * }}}
    * As you can see, tuples can be created easily:
    */
  def oneIndexedTuples(res0: String, res1: String) {
    val tuple = ("apple", "dog")
    val fruit = tuple._1
    val animal = tuple._2

    fruit should be(res0)
    animal should be(res1)
  }
  // ...
}
```

andyscott:
This fixes the formatting bug for certain code blobs in exercise comments. It should close #309.

### DIFF
--- a/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala
+++ b/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/comments.scala
@@ -299,9 +299,18 @@ private[compiler] object CommentRendering {
   }
 
   private[this] def formatCode(code: String): String = {
-    def wrap(code: String): String = s"""object Wrapper { $code }"""
+    def wrap(code: String): String = s"""// format: OFF
+      |object Wrapper {
+      |// format: ON
+      |$code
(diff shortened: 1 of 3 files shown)


dialelo:
itworks

Is fixed now, great work!

Issue #332, opened 2016-04-01
Compilation error: Unable to evaluate arg Option(1) as scala.Option[scala.Int]


raulraja:
This is still happening in all libraries. The evaluator is somehow broken and yields inconsistent evaluations after the imports where auto discovered.

I think for the time being we just need to specify per section which imports will be provided for user input to be evaluated against.

This is a high priority as it prevents from evaluating exercises.

![sc](https://cloud.githubusercontent.com/assets/456796/14203732/08dafa4c-f800-11e5-95c7-bb517c2c73f0.png)

raulraja:
@andyscott Was this already taken care of? let me know so we can close this issue. thanks!

andyscott:
I believe it was addressed-- we should double check before closing this!

raulraja:
I can't repro it so we'll reopen it if it comes back.

Issue #167, opened 2016-02-10
Shapeless exercises


FPerezP:
We need a set of exercises about Shapeless so we can show it on our Presentation Demo in TypeLevel Summit.

Having a set of 5 exercises is fair enough for the demo, we'll add more exercises in the future.

raulraja:
Currently blocked by #182

andyscott:
This should no longer be blocked by #182.

raulraja:
This was completed though more exercises will be welcome. Will create separate issues for those.

PR #333, opened 2016-04-01
More shapeless exercises #167

raulraja:
The following exercises cover most of the content in the Shapeless [Feature Overview Guide](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0).

Most of them are adapted from the context of the guide but perhaps they need to be revisited to provide better exercise examples and adding missing cases in each one of the sections.

I have not been able to test all the exercises because of #332 but given we need a full pass in all content for all the 3 existing libraries I think this is good to be merged for now and we can address any UI related fixes or typos as a general pass.

@scala-exercises/core-mergers Please review. Thanks!


site/content/src/main/scala/shapeless/AutoTypeClassExercises.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+}
+
+/** == Automatic type class instance derivation ==
+  *
+  * Based on and extending `Generic` and `LabelledGeneric`, Lars Hupel ([[https://twitter.com/larsr_h @larsr_h]]) has contributed the `TypeClass`
+  * family of type classes, which provide automatic type class derivation facilities roughly equivalent to those available
+  * with GHC as described in [[http://dreixel.net/research/pdf/gdmh.pdf "A Generic Deriving Mechanism for Haskell"]].  There is a description of an
+  * earlier iteration of the Scala mechanism [[http://typelevel.org/blog/2013/06/24/deriving-instances-1.html here]], and examples of its use deriving `Show` and `Monoid`
+  * instances [[https://github.com/milessabin/shapeless/blob/master/examples/src/main/scala/shapeless/examples/shows.scala here]]
```
dialelo: We can make these links independent of Shapeless' `master` branch because those files may be deleted. This can be done visiting the URL and pressing `y`, it will redirect you to a URL with the hash of the blob you are looking at like https://github.com/milessabin/shapeless/blob/f1d763732ac56fed1a811bc0f301d96f3c81bb7b/examples/src/main/scala/shapeless/examples/shows.scala

This way the links will not break and always point at what we were looking at when writing this.

site/content/src/main/scala/shapeless/SizedExercises.scala
```diff
@@ -0,0 +1,63 @@
...
+      Sized("Types and Programming Languages", "Benjamin Pierce"),
+      Sized("The Implementation of Functional Programming Languages", "Simon Peyton-Jones")
+    )
+  }
+
+  import Helper._
+
+  /** In the example below we define a method `csv` whose signature guarantees at compile time that there are exactly as many
+    * column headers provided as colums
+    * TODO : what would be a good exercise for stuff that can only be proven at compile time???
```
dialelo: Maybe we want to extract this comment outside the exercise test so it doesn't get rendered in the UI.

dialelo:
Awesome content, can't wait to go through all of these exercises. Besides of the minor nitpicks LGTM!

site/content/src/main/scala/shapeless/AutoTypeClassExercises.scala
raulraja: I kept them as is since that is what they are in the actual shapeless Feature Overview which targets shapeless 2.0.0. Not sure if Miles intention was to point to the latest or the ones that work with that version of the guide.

site/content/src/main/scala/shapeless/SizedExercises.scala
raulraja: This exercise as it stands can't be completed in the UI and that is why I'm just having the user type a 1. I'd rather keep this comment around so we don't forget to get rid of it or rewrite it. Alternatively I'd rewrite the section altogether but I think once we ask Miles for feedback he can come with a good exercise to test this area.

dialelo: Excellent.

site/content/src/main/scala/shapeless/AutoTypeClassExercises.scala
dialelo: :+1:

PR #342, opened 2016-04-26
Installation instructions in the README

dialelo:
I've written the instructions based on the great work by @eliasdefrancisco but put it on the README and removed some level of detail. All feedback is welcome.

You can preview it [here](https://github.com/scala-exercises/scala-exercises/tree/al-how-to-install#getting-started).


rafaparadela:
Looks good to me.

PR #343, opened 2016-05-01
Fix wrong parameter of "psql" when creating the db

studiodev:
From the documentation, the parameter for running a command is "-c", not "-C" (http://www.postgresql.org/docs/9.4/static/app-psql.html)


dialelo:
Great catch, thanks!

PR #348, opened 2016-05-18
Implement Evaluator based on Scala Compiler

MasseGuillaume:
* Evaluator produce an EvalResult[T] it has more information than the original EvaluationResult[T]
  double check to see what can be bubble back to the ui
    * RangePosition in CompilationInfo would need to be shifted (see wrapedCode in eval[T])
* I'm using sbt BuildInfo in ExerciceService to find the correct classpath (artifacts). This would changed when deployed.
  I'm not too familiar with Play dependency injection, it's up to you guys to figure it out.
* MethodEval security (jvm SecurityManager) is off by default in ExercisesService. Play and sbt are running on the same jvm. It should work fine when deployed.

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -38,11 +38,18 @@ lazy val compiler = (project in file("compiler"))
   .dependsOn(runtime % "test")
 
 // ~ Exercise Runtime
-lazy val runtime = (project in file("runtime"))
+lazy val `runtime-example` = project
   .settings(
-    name            := "runtime"
+    name := "runtime-example"
   )
+  .settings(libraryDependencies += Dep.cats.core)
+
+lazy val runtime = (project in file("runtime"))
   .settings(commonSettings: _*)
+  .settings(
+    name := "runtime"
+  )
+  .dependsOn(`runtime-example` % "test")
   .settings(libraryDependencies <++= scalaVersion(scalaVersion =>
     compilelibs(
       Dep.scala.compiler(scalaVersion),
--- a/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/Compiler.scala
+++ b/core/compiler/src/main/scala/com/fortysevendeg/exercises/compiler/Compiler.scala
@@ -54,6 +54,7 @@ case class Compiler() {
       comment:         RenderedComment.Aux[Mode.Exercise],
       code:            String,
       qualifiedMethod: String,
+      packageName:     String,
       imports:         List[String]                       = Nil
     )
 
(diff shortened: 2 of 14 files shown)


raulraja:
@MasseGuillaume Great stuff. I'll defer to @andyscott review of this PR given he had spent the most time around the evaluation code prior to this change.

MasseGuillaume:
@andyscott ping ?

dialelo:
It seems like the build failed due to an [unresolved dependency](https://travis-ci.org/scala-exercises/scala-exercises/builds/131005108#L1958), can you take a look @MasseGuillaume ?

MasseGuillaume:
@dialelo I solved the issue with the `runtime-example` resolution. Now it's something else: https://travis-ci.org/scala-exercises/scala-exercises/builds/132277148#L1089

MasseGuillaume:
@dialelo done ;-)

dialelo:
We may want to review the error messages in the future but this is awesome!

timeout

I've tested it locally and works perfectly, LGTM :shipit:

dialelo:
Hmm, it seems like it doesn't find some imports. In the `Apply` section of cats we are importing an object defined in the same package called `ApplyHelper` but is not available when evaluating the code:

notfoundimport

dialelo:
Same problem for every section that imports an object defined on its own package, any ideas about what can be failing @MasseGuillaume? We can work around it including all the code in the section file but is less than ideal.

MasseGuillaume:
@dialelo I'm checking this out ;-)

dialelo:
Awesome :sparkles:

MasseGuillaume:
An import is missing from the Exercice plugin

This is the generated code:
```
import cats.syntax.cartesian._
import cats.syntax.apply._
import cats.std.all._
import cats._
import ApplyHelpers._
import org.scalatest._
class Eval2803112120684302132 extends (() ⇒ Any) {
  def apply() = catslib.ApplySection.applyExtendsFunctor(Option("1"), Option(2), None)
}
```

it's missing `package catslib` or `import catslib._`

dialelo:
@MasseGuillaume can you review and merge [my PR to your master branch](https://github.com/MasseGuillaume/scala-exercises/pull/1) so it is included in this PR, please? Thanks in advance!

MasseGuillaume:
@dialelo tested locally with your PR works perfectly. Thanks!

dialelo:
@rafaparadela @raulraja @andyscott please take a look and let us know if you run into any problems. If not I think this is good to merge.

andyscott:
I will make time to look at this ASAP. Sorry for the delay guys.

MasseGuillaume:
Also, I took some bits from Twitter Eval
https://github.com/twitter/util/blob/develop/util-eval/src/main/scala/com/twitter/util/Eval.scala#L74-L86
https://github.com/twitter/util/blob/develop/util-eval/src/main/scala/com/twitter/util/Eval.scala#L319-L365
It's to extract useful Path from the classpath.

It's an Apache 2.0 so it's compatible. We need to comply with the [Redistribution section](https://github.com/twitter/util/blob/develop/LICENSE#L90-L129)

PR #352, opened 2016-05-24
Removed unprincipled exercises and those implemented with mutability

raulraja:
The following PR gets rid of exercises that either use mutability or showed Scala techniques that may not be the best for newcomers getting introduced to the lang.
@dialelo @rafaparadela Can you please review? thanks!


site/content/src/main/scala/stdlib/Implicits.scala
```diff
@@ -102,12 +102,9 @@ object Implicits extends FlatSpec with Matchers with exercise.Section {
   def asDefaultImplicits(res0: Float, res1: Float) {
```
dialelo: if we no longer use `res1` in the body of the exercise we can get rid of it.

site/content/src/main/scala/stdlib/Implicits.scala
```diff
@@ -116,13 +113,10 @@ object Implicits extends FlatSpec with Matchers with exercise.Section {
     def howMuchCanIMake_?(hours: Int)(implicit amount: BigDecimal, currencyName: String) =
       (amount * hours).toString() + " " + currencyName
 
-    implicit var hourlyRate = BigDecimal(34.00)
```
dialelo: :fire: `res1`

site/content/src/main/scala/stdlib/Lists.scala
```diff
@@ -48,11 +48,12 @@ object Lists extends FlatSpec with Matchers with exercise.Section {
     a should equal(List(res0, res1, res2))
   }
 
-  /** Lists can be accessed via head and tail
+  /** Lists can be accessed via head, headOption and tail.
+    * Accessing List via `head` is unsafe and may result in a IndexOutOfBoundsException
     */
   def headAndTailLists(res0: Int, res1: Int, res2: Int) {
     val a = List(1, 2, 3)
-    a.head should equal(res0)
```
dialelo: we may want to encourage `Option` instead of `Some` for constructing optional values.

dialelo:
Awesome changes. I noticed some TODOs, are we addressing them as part of this pull request? I can create the appropiate issues if we're going to do it after merging this.

raulraja:
@dialelo Addressed all your comments. RE the TODOs I have already added the issues. Once is related to the compiler not being able to process multiline string literals and the other was related to the contravariance exercises which need to be rewritten without mutability.

dialelo:
Awesome, LGTM :shipit:

PR #357, opened 2016-06-03
Extract core

dialelo:
This PR extracts the core dependencies, which now live in separate repositories, and puts the site project at the root.

@raulraja @rafaparadela please take a look.


rafaparadela:
LGTM

PR #359, opened 2016-06-06
Group projects

dialelo:
@raulraja @rafaparadela there is one problem with adding `content` here since it requires the `sbt-exercise` plugin, which is one of the subprojects. I tried adding the compiler plugin only to the subproject but plugins are global to all projects; should we keep `content` separate? Another alternative is to have it as a subdir in the repository but not as a subproject, and have its own `build.sbt`. Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,9 @@ env:
     SBT_OPTS="-XX:+UseConcMarkSweepGC -XX:MaxPermSize=512m"
 install:
   - gem install sass
+script:
+  - sbt 'server/test'
+  - sbt 'client/test'
+  - sbt 'definitions/test'
+  - sbt 'runtime/test'
+  - sbt 'compiler/test'
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
   - gem install sass
 script:
   - sbt 'publishAll'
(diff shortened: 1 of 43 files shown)


raulraja:
Let's keep it a separate project since ultimately each lib featured in scala exercises will have it's own project. Inmediately after combining the repos we need to also split content and separate the stdlib, cats and shapeless into independent repos. This will help down the road separate their transitive dependencies so they run in their own classloader to avoid version collisions among libraries and projects they use cats, shapeless , etc....

dialelo:
:+1:, let's make the build green before merging this.

dialelo:
I ignored some ill-behaved tests. Due to having to generate many case classes and discarding a lot of data Scalacheck sometimes ends up failing a property, even after a few successful evaluations. It seems like Scalacheck doesn't respect the Scalacheck config I give it, so i'll convert those tests to plain example-based tests in a future PR.

dialelo:
@raulraja @rafaparadela please take a look and merge at your convenience.

rafaparadela:
According to our purpose of having multiple projects in this repo again, all these changes look good to me.

dialelo:
@raulraja please take a look.

raulraja:
LGTM

PR #360, opened 2016-06-06
Update content dependencies

dialelo:
Now that we have split the content into their own repositories, we need to update the site's dependencies accordingly.


dialelo:
@rafaparadela plz take a look.

rafaparadela:
Great 👏  
LGTM, thank you @dialelo

PR #363, opened 2016-06-06
Adds settings for distributing projects in Sonatype

rafaparadela:
This PR adds setting for distribution of projects on Sonatype.

@dialelo @raulraja Could you take a look please? Thank you.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,10 +46,10 @@ lazy val server = (project in file("server"))
       evolutions,
       cache,
     ws,
-      "org.scala-exercises" %% "runtime" % "0.0.1" changing(),
-      "org.scala-exercises" %% "content-stdlib" % "0.0.1",
-      "org.scala-exercises" %% "content-cats" % "0.0.1",
-      "org.scala-exercises" %% "content-shapeless" % "0.0.1",
+      "org.scala-exercises" %% "runtime" % "0.0.0-SNAPSHOT" changing(),
+      "org.scala-exercises" %% "content-stdlib" % "0.0.0-SNAPSHOT",
+      "org.scala-exercises" %% "content-cats" % "0.0.0-SNAPSHOT",
+      "org.scala-exercises" %% "content-shapeless" % "0.0.0-SNAPSHOT",
       "org.slf4j" % "slf4j-nop" % "1.6.4",
       "org.postgresql" % "postgresql" % "9.3-1102-jdbc41",
       "com.vmunier" %% "play-scalajs-scripts" % "0.2.1",
@@ -125,7 +125,7 @@ lazy val definitions = (project in file("definitions"))
   .settings(publishSettings:_*)
   .settings(
     organization := "org.scala-exercises",
-    version := "0.0.1",
+    version := "0.0.0-SNAPSHOT",
     name := "definitions",
     scalaVersion := "2.11.7",
     libraryDependencies ++= Seq(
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -40,17 +40,16 @@ lazy val server = (project in file("server"))
...
       filters,
       jdbc,
       evolutions,
       cache,
     ws,
-      "org.scalaexercises" %% "runtime" % "0.0.0-SNAPSHOT" changing(),
-      "org.scalaexercises" %% "content-stdlib" % "0.0.0-SNAPSHOT",
-      "org.scalaexercises" %% "content-cats" % "0.0.0-SNAPSHOT",
-      "org.scalaexercises" %% "content-shapeless" % "0.0.0-SNAPSHOT",
+      "org.scala-exercises" %% "runtime" % "0.0.1" changing(),
```
raulraja: These shall remain snapshots until we have a final version because we want up to date snapshots of master.

build.sbt
```diff
@@ -217,6 +220,45 @@ lazy val `sbt-exercise` = (project in file("sbt-exercise"))
...
 
+// Distribution
+
+lazy val gpgFolder = sys.env.getOrElse("SE_GPG_FOLDER", ".")
+
+lazy val publishSettings = Seq(
+  organizationName := "Scala Exercises",
+  organizationHomepage := Some(new URL("http://scala-exercises.org")),
+  startYear := Some(2016),
+  description := "Scala Exercises: The path to enlightenment",
```
raulraja: We need a better tagline. Maureen may be of help here.

build.sbt
```diff
@@ -217,6 +220,45 @@ lazy val `sbt-exercise` = (project in file("sbt-exercise"))
...
+  homepage := Some(url("http://scala-exercises.org")),
+  pgpPassphrase := Some(sys.env.getOrElse("SE_GPG_PASSPHRASE", "").toCharArray),
+  pgpPublicRing := file(s"$gpgFolder/pubring.gpg"),
+  pgpSecretRing := file(s"$gpgFolder/secring.gpg"),
+  credentials += Credentials("Sonatype Nexus Repository Manager",  "oss.sonatype.org",  sys.env.getOrElse("PUBLISH_USERNAME", ""),  sys.env.getOrElse("PUBLISH_PASSWORD", "")),
+  scmInfo := Some(ScmInfo(url("https://github.com/scala-exercises/site"), "https://github.com/scala-exercises/site.git")),
+  licenses := Seq("Apache License, Version 2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0.txt")),
+  publishMavenStyle := true,
+  publishArtifact in Test := false,
+  pomIncludeRepository := Function.const(false),
```
raulraja: What is Function.const?

rafaparadela: Given that  `pomIncludeRepository` expect a function that returns a boolean, I used  `Function.const` which represents the constant function:

```scala
def const[T, U](x: T)(y: U): T = x
```

Actually we can express it thus:

```scala
pomIncludeRepository := { _ => false }
```

build.sbt
rafaparadela: Makes sense

build.sbt
rafaparadela: 👍

build.sbt
raulraja: No,  it's great, just I have never seen that  before.

raulraja:
LGTM!  tagline can be separately addressed.

PR #365, opened 2016-06-07
Disable automatic running of evolutions on app load

dialelo:
@raulraja @rafaparadela plz take a look, this will avoid running evolutions automatically in production. During development, Play will detect schema changes and prompt you to apply them when running the app.

### DIFF
--- a/server/app/com/fortysevendeg/exercises/ApplicationLoader.scala
+++ b/server/app/com/fortysevendeg/exercises/ApplicationLoader.scala
@@ -12,7 +12,6 @@ import doobie.contrib.hikari.hikaritransactor.HikariTransactor
 import doobie.util.transactor.{ DataSourceTransactor, Transactor }
 import play.api.ApplicationLoader.Context
 import play.api._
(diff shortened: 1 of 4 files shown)


server/app/com/fortysevendeg/exercises/ApplicationLoader.scala
```diff
@@ -62,10 +60,6 @@ class Components(context: Context)
 
   implicit val wsClient: WSClient = NingWSClient()
 
-  // touch lazy val to enable
-  applicationEvolutions
-  override def dynamicEvolutions: DynamicEvolutions = new DynamicEvolutions
```
raulraja: Can we check here if the app is in DEV mode and then run them only in such case?

raulraja: Nevermind, I guess the Evolutions UI takes care of that as discussed offline.

raulraja:
LGTM

dialelo:
I disabled automatic evolutions in Play configuration too; in DEV mode, evolutions will be automatically applied.

raulraja:
LGTM again! :+1:

PR #387, opened 2016-06-09
Distinguish between not found libraries, sections and server errors

dialelo:
@raulraja @rafaparadela plz take a look.


raulraja:
LGTM

PR #397, opened 2016-06-13
Retrieve a valid access token for Github requests

### LINKED ISSUE #396: Avoid rate limit when contributions are fetched
Even now when contributions are retrieved during compiling process, we should provide an access token in order to avoid:

```
github4s.GithubResponses$UnexpectedException: Failed invoking get with status : 403, body : 
 {"message":"API rate limit exceeded for 84.122.87.254. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

Because authenticated requests have a higher rate limit.

rafaparadela:
This PR fixes #396 . It get an environment variable called `GITHUB_TOKEN` in order to avoid the rate limit exception.

So, you should [create a new personal access token with `repo` scope](https://github.com/settings/tokens/new) and create an env var called `GITHUB_TOKEN` with it.

@dialelo Could you review please? Thank you.


dialelo:
:+1:

PR #402, opened 2016-06-13
Rename database url env var

rafaparadela:
DO NOT MERGE!

It seems like Heroku create an environment variable with datasource data, called `DATABASE_URL` for review apps. Let's try if they can be properly deployed if we retrieve this info to instantiate data source.


rafaparadela:
Test concluded

PR #406, opened 2016-06-13
Accumulate imports in order

dialelo:
Fixes a bug we found in the compiler: emitting the imports in reverse order. @raulraja @rafaparadela please take a look.


raulraja:
LGTM!

PR #414, opened 2016-06-15
Enable New Relic

raulraja:
@dialelo Since I have to test this directly in the prod heroku app this may take multiple reviews. Just a heads up. thanks.


dialelo:
:+1:

Issue #410, opened 2016-06-14
Loading latest uncompleted section


raulraja:
When entering a Library content and loading the section currently we are redirecting to the first session in the list.

Ideally the user will land in the last uncompleted section so he can continue were he left off.
Alternatively we could store the user last visited section and load that one.
The first section should always be loaded in the case that the user has not progress point.

PR #420, opened 2016-06-15
Go to last seen section when visiting a library

dialelo:
Addresses #410. I don't currently check if the last seen section is completed so it will show completed sections anyway. I think it makes more sense and won't confuse users.

When we implement the dialog that takes you to the next section after the current is finished, a user revisiting the last seen section will be prompted to go the next if such section is completed. So I think is OK to take you to the last seen section regardless if is completed, thoughts @raulraja ?


raulraja:
LGTM!

PR #434, opened 2016-06-16
Add metadata

### LINKED ISSUE #429: Meta content missing
Currently sharing the site in social networks causes no preview or rich media. We need appropriate meta tags for twitter, facebook, etc.

rafaparadela:
This PR closes #429. It adds metadata in home/library views

@dialelo @raulraja Could you review please? Thank you


rafaparadela:
Note that I'm creating two new issues for:

- Asking Maureen for a new tagline/description.
- Asking Isra for a better poster image.

server/app/views/templates/base/head.scala.html
```diff
@@ -0,0 +1,33 @@
...
+
+<meta charset="UTF-8">
+<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<link rel="shortcut icon" type="image/png" href="@routes.Assets.at("images/favicon.png")" />
+
+<title>@title</title>
+<meta name="description" content="@description">
+<meta name="author" content="47 Degrees">
+<meta property="og:image" content="@routes.Assets.at("images/poster.png")"/>
```
raulraja: Would this result in a relative URL? If so I think we need absolute URLs in all instances.

dialelo:
Apart from Raúl's comment LGTM :+1:

server/app/views/templates/base/head.scala.html
```diff
@@ -0,0 +1,33 @@
...
+<meta property="og:title" content="@title"/>
+<meta property="og:site_name" content="@current.configuration.getString("application.name").getOrElse("Scala Exercises")"/>
+<meta property="og:url" content="@current.configuration.getString("application.url").getOrElse("https://www.scala-exercises.org")"/>
```
raulraja: This is not the URL we want to share. If someone shares a section this URL is always going to be the main site and not the section. We need to use the ReverseController of the given action to bring the current absolute URL here.

rafaparadela:
Addressed comments:
- Now the URL of poster image is expressed as absolute path.
- Now metadata `url` is sensible to the current path, the target link is consistently to the shared section.

@raulraja Is good to be merged?

raulraja:
LGTM

PR #448, opened 2016-06-20
Add Andy

raulraja:
When the list of developers was added @andyscott was left out. 
This an undeniably an oversight since Andy is entirely the author of the Exercise compiler.
It also orders the list alphabetically in order to not imply any hierarchy in the importance of contributions.

Apologies @andyscott 

@rafaparadela can you please review?


rafaparadela:
LGTM. Sorry about that @andyscott

PR #449, opened 2016-06-21
428 stars watchers forks

### LINKED ISSUE #428: Re-enable stars, watchers and forks in the jumbotron
The current jumbotron in the site is missing the call to action to receive love from users. Please reenable in new site right below the tagline.

rafaparadela:
This PR fixes #428 

- It retrieves the repository data.
- It shows stars, forks and watchers in Home.

Preview: 

http://scala.run/28ReoWM

@raulraja @dialelo Could you take a look please? Thank you.


andyscott:
Maybe squash all the merges into one commit?

dialelo:
:+1: Are we doing authenticated calls to GitHub? I'm worried about rate limiting.

In the future we could move this request to be performed in the client but as long as we don't get rate limited we can make the requests in the server for the moment.

rafaparadela:
@dialelo Yes, we are requesting the repo data with an authenticated call to avoid the rate limit. Given that we are retrieving the environment variable for others Github calls, I decided to re-use it as well:

https://github.com/scala-exercises/scala-exercises/pull/449/files#diff-7ed401ce2d238c772383c0972c539c5aR110

Makes sense?

dialelo:
Nice, makes sense :+1: LGTM

server/app/views/templates/home/index.scala.html
```diff
@@ -44,6 +45,31 @@
           <div class="col-md-8 col-md-offset-2">
             <h1>@current.configuration.getString("application.tagline").getOrElse("")</h1>
             <p>@current.configuration.getString("application.description").getOrElse("")</p>
+
+            <div class="github-details">
+
+              <div class="row">
+                <div class="col-xs-4">
+                  <a target="_blank" href="https://github.com/47deg/scala-exercises/watchers">
+                    <h3><i class="fa fa-eye"></i> @repo.status.subscribers_count</h3>
```
raulraja: Why the space before the rendered value?

raulraja:
This is going to be an issue with rate limiting specially in the home page. We need to cache these values and only have them refreshed every so often. for example rerender them once every 30 mins. Aomething in these lines but adapted to Future.
```scala
val repo = cache.getOrElse[User]("home.repo.info", duration) {
  GH.getRepo(owner, repo)
}
```
http://stackoverflow.com/questions/34612363/cache-getorelse-on-futures-in-playframework

server/app/views/templates/home/index.scala.html
rafaparadela: Because we need a space between the eye icon and the number.

PR #453, opened 2016-06-23
Highlight code blocks in the client side

dialelo:
We removed the CSS for the highlighting accidentaly, causing the highlighting not to work. I've restored the CSS style linking to the theme we were using. There is [a great amount of themes available](https://highlightjs.org/static/demo/) so we may want to consider changing it at some point or even make it a config option in the application. @raulraja please take a look.


raulraja:
LGTM!

Issue #319, opened 2016-03-29
Incorrect formatting for some exercises


dialelo:
This is how a method in the Monoid section looks like:

```scala
  /**
    * To use this
    * with a function that produces a tuple, we can define a `Monoid` for a tuple 
    * that will be valid for any tuple where the types it contains also have a 
    * `Monoid` available.
    *
    * This way we are able to combine both values in one pass, hurrah!
    */
  def tupleMonoid(res0: (Int, String)) = {
    implicit def tupleMonoid[A : Monoid, B : Monoid]: Monoid[(A, B)] =
      new Monoid[(A, B)] {
        def combine(x: (A, B), y: (A, B)): (A, B) = {
          val (xa, xb) = x
          val (ya, yb) = y
          (Monoid[A].combine(xa, ya), Monoid[B].combine(xb, yb))
        }
        def empty: (A, B) = (Monoid[A].empty, Monoid[B].empty)
      }

    val l = List(1, 2, 3, 4, 5)
    l.foldMap(i => (i, i.toString)) should be(res0)
  }
```

It defines a implicit monoid instance before using it. However, it gets rendered like this:

![implicit](https://cloud.githubusercontent.com/assets/409039/14104666/999428f8-f5a6-11e5-9cb3-535d2ba76431.png)

The `implicit` keyword is stripped and the next statements are not correctly indented.

andyscott:
That's really odd! I will take a look.

raulraja:
Is this still an issue or can it be closed?

dialelo:
Still an issue.

dialelo:
Here's another instance of formatting problems even when not using implicits, in Shapeless' Sized section:

![formatting](https://cloud.githubusercontent.com/assets/409039/16040366/60687f46-322f-11e6-978c-f6899d4783a0.png)

dialelo:
Note that the above example is not an exercise body but sample code. We should also format sample code the same way we do with the exercise code.

PR #454, opened 2016-06-23
Correctly format sample code blocks and exercise code

dialelo:
- [ ] Merge #453 first

Partially addresses the issues outlined in #319, now both exercises and sample code are formatted correctly. The problem with the implicit keyword in #319 happens during the code extraction, so I'm opening a separate issue to address it.

@raulraja please take a look.


raulraja:
LGTM!

PR #482, opened 2016-06-28
Redirect loop fix

raulraja:
Looks into the x-forward-proto headers to determine if https since behind the Heroku load balancer request.secure returns false.
@dialelo Can you review? thanks!


dialelo:
:+1: LGTM

Issue #490, opened 2016-06-29
AbstractMethodError when evaluating exercises


raulraja:
Compilation error: java.lang.AbstractMethodError: Eval9055044378944555573.apply()Ljava/lang/Object;

dialelo:
@raulraja is fixed now, i'll close this issue when somebody other than me confirms it. @FPerezP can you take a look?

PR #491, opened 2016-06-29
Version the content and site together

dialelo:
For avoiding accidentaly introduced incompatibilities between the
content, runtime and the site I've decided to not continue using wildcard
version specs like 0.1.+. This means that each time we want to publish
the content we'll have to bump the content repos *and* the site, but is
an extra check that can avoid problems introduced by automatic
deployments like in #490

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -81,9 +81,10 @@ lazy val server = (project in file("server"))
       evolutions,
       cache,
     ws,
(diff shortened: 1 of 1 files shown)


dialelo:
@raulraja let me know your thoughts.

raulraja:
Makes perfect sense. LGTM!

PR #504, opened 2016-07-04
Bump 0.1.2 for depending on non-snapshot github4s

dialelo:
The 0.1.1 artifact depends on github4s@0.5-SNAPSHOT but since there is a proper release the snapshot version is no longer available. Bumping a new release for dependending on github4s@0.5.

@raulraja @franciscodr please take a look.


raulraja:
LGTM!

PR #512, opened 2016-07-05
Fixes bug when the last seen section was renamed or removed

### LINKED ISSUE #507: Take into account that database section names can get out of sync with runtime section names
We currently use the name of a section as part of the identity of the evaluation of an exercise, and store it together with the user input whenever they evaluate an exercise. 

This is a problem since section names can change or dissapear entirely so we should handle the cases when we have a section in the database that is not available at runtime. This involves

- creating a migration every time a section identity that's been published changes
- handling the scenario outlined above in our code, specifically when visiting a library and resolving the last seen section from the user's progress.

franciscodr:
This pull request closes #507. It solves the found bug  when the last seen section was renamed or removed.

@dialelo @raulraja Could you take a look when you have a chance? Thanks


dialelo:
:+1: LGTM

PR #519, opened 2016-07-06
Depend on runtime through library dependency instead of sbt

### LINKED ISSUE #518: Exercise libraries not appearing in the local development environment
@jdesiloniz  and @juanpedromoreno have experienced this problem locally, i'm investigating the causes.

dialelo:
Closes #518 

@jdesiloniz @juanpedromoreno can you confirm that fixes the issue for you? Thanks in advance!


juanpedromoreno:
It's working now :shipit: 

Code looks good BTW.
Thanks @dialelo !

PR #531, opened 2016-07-07
Allow evaluation of exercises even after being solved

### LINKED ISSUE #459: Exercise evaluation is not done if a previously evaluated answer is used again
If you answer an exercise, change it and the go back to any previous answer the exercise is not re-evaluated. Using the first exercise in the stdlib as an example:

true should be(___)

Repro steps:

1. Type 'true' in the exercise blank and hit Enter
2. The exercise is evaluated as correct and highlighted green
3. Clear your previous answer and type 'false' and hit Enter
4. The exercise is evaluated as incorrect and failed assertion shown
5. Clear your previous answer and type 'true' and hit Enter
6. The exercise is not evaluated

Expected result:
The exercise should be re-evaluated even if the user typed that answer in past evaluations.

Appears to be a client side issue as refreshing the page clears the state and you can re-evaluate the exercise with a past answer.

dialelo:
Closes #459

@raulraja please take a look.

### DIFF
--- a/client/src/main/scala/Model.scala
+++ b/client/src/main/scala/Model.scala
@@ -26,7 +26,7 @@ case class ClientExercise(
 
   def isBeingEvaluated: Boolean = state == Evaluating
 
-  def canBeCompiled: Boolean = isFilled && !isSolved && !isBeingEvaluated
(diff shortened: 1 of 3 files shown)


raulraja:
LGTM!

PR #532, opened 2016-07-07
Creates different evaluator instances each time

juanpedromoreno:
Please, @raulraja Review this change at your convenience.

Thanks.


dialelo:
:+1:

raulraja:
@dialelo since we no longer aggregate runtime.... and we are including a hard dependency to a remote artifact for the runtime would any of this code actually make it to Heroku or do we have to manually publish it? I have been commiting to @juanpedromoreno branch another version of the evaluator with Twitter Eval and I'm not seeing any of my println statements in the preview app :(

PR #539, opened 2016-07-08
Capture non fatal errors in exercise evaluation

dialelo:
@raulraja @juanpedromoreno please take a look.


juanpedromoreno:
LGTM!

Issue #540, opened 2016-07-08
Don't let people inject evil code


anadoba:
I was actually learning, but my friend came up and said `Why don't we try putting some code there?`

so when the exercise looked roughly like this

```
val some = Some(123)

some.get shouldEqual _____    <- type something here
```

I put
`{ while (true) {}; 123 }`

Unfortunately server went down while testing 👯 
Looks like the whole website is down on `Application Error` title and `404` page.
I'm sorry :)

dialelo:
It's because yesterday we replaced the problematic evaluation code with Twitter's eval and we are not respecting timeouts now, working to fix it. Thanks for reporting :blush:

PR #541, opened 2016-07-08
Evaluate code with the configured timeout

dialelo:
Solved #540 

@raulraja please take a look.

### DIFF
--- a/runtime/src/main/scala/org/scalaexercises/exercises/Evaluator.scala
+++ b/runtime/src/main/scala/org/scalaexercises/exercises/Evaluator.scala
@@ -21,9 +21,8 @@ import java.util.concurrent.{ TimeoutException, Callable, FutureTask, TimeUnit }
 
 import scala.util.Try
 import scala.util.control.NonFatal
-import scala.concurrent._
 import scala.concurrent.duration._
-import scala.concurrent.ExecutionContext
+import scala.concurrent._
 import scala.language.reflectiveCalls
 
 import com.twitter.util.Eval
--- a/runtime/src/main/scala/org/scalaexercises/exercises/Evaluator.scala
+++ b/runtime/src/main/scala/org/scalaexercises/exercises/Evaluator.scala
@@ -21,9 +21,14 @@ import java.util.concurrent.{ TimeoutException, Callable, FutureTask, TimeUnit }
 
 import scala.util.Try
 import scala.util.control.NonFatal
+import scala.concurrent._
 import scala.concurrent.duration._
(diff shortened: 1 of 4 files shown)


raulraja:
LGTM but let's review the previous evaluator code to see if there were any additional security guards that we may be have missed last night when we rush to fix the threading issues in the old evaluator.

runtime/src/main/scala/org/scalaexercises/exercises/Evaluator.scala
```diff
@@ -97,4 +99,15 @@ class Evaluator(timeout: Duration = 20.seconds) {
...
+
+  def apply[T](pre: String, code: String): EvalResult[T] = {
+    val allCode = s"""
+                     |$pre
+                     |$code
+                   """.stripMargin
+    val fut = Future({ eval(allCode) })
+    Try(
+      Await.result(fut, timeout)
+    ).getOrElse(EvalResult.Timeout())
```
maasg: With this construction, I think we might be loosing any failure information from the execution of `eval(...)` and replacing it by a timeout. Don't know the codebase to tell if that should be the intention.

dialelo: Good point! The `EvalResult` returned by the `Future` already carries an exception if it occurs during the execution of the code, the `Await.result` will throw an exception only for the timeouts.

runtime/src/test/scala/com/fortysevendeg/exercises/EvaluatorSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import scala.concurrent.duration._
+import org.scalatest._
+
+class EvaluatorSpec extends FunSpec with Matchers {
+
+  describe("evaluation") {
+    it("fails with a timeout when takes longer than the configured timeout") {
+      val evaluator = new Evaluator(1 second)
+      val result: EvalResult[Int] = evaluator("", "{ while(true) {}; 123 }")
```
maasg: what I meant above:
```
evaluator("", "throw new Exception("bum")") should be ______?
```

dialelo: In that case it'd be `EvalRuntimeError` containing the thrown exception, which is of type `EvalResult[T]`. The original exception is captured and its message shown to the user in the UI.

raulraja:
LGTM, let's address the security concerns in a different PR

PR #551, opened 2016-07-11
JP - SE-510 DRY Test.scala

### LINKED ISSUE #510: DRY Test.scala util used in all exercises modules
https://github.com/scala-exercises/exercises-cats/blob/master/src/test/scala/exercises/Test.scala is copied and pasted everywhere and should be placed in a common place where it can be reused as a dependency.

juanpedromoreno:
Fixes #510 

Please, @dialelo @raulraja could you review?

Thanks!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ def wartSettings =
 
 lazy val commonSettings = Seq(
   organization := "org.scala-exercises",
-  version := "0.2.1-SNAPSHOT",
+  version := "0.2.2-SNAPSHOT",
   scalacOptions ++= Seq(
     "-deprecation",
     "-encoding",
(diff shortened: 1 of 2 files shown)


raulraja:
LGTM

build.sbt
```diff
@@ -43,7 +43,7 @@ def wartSettings =
 
 lazy val commonSettings = Seq(
   organization := "org.scala-exercises",
-  version := "0.2.1-SNAPSHOT",
+  version := "0.2.2-SNAPSHOT",
```
fedefernandez: Just a question, can we use the same version? I think we haven't published the `0.2.1` version

juanpedromoreno: This PR will generate a new snapshot version. There are other PRs like https://github.com/scala-exercises/exercises-cats/pull/34 which are increasing the version. We have to do this for every module once 0.2.2 snapshot is released. In the meantime, you can use the 0.2.1.

Does it answer your question? Thaks for asking @fedefernandez

fedefernandez: Yes, thanks. My main concern is that we're skipping the `0.2.1` final (not snapshot) version, but if the team agrees I'm fine with that.

dialelo:
:+1:

raulraja:
LGTM, merge at will

PR #559, opened 2016-07-13
Prevents empty sections list for Library objects

### LINKED ISSUE #511: The list of sections in the library definitions should be a NonEmptyList
The list of section is currently required to be a `List[Section]`, which can be empty. Since we don't support libraries withouth sections we should make this invariant explicit in the type of sections. This will require modifying the `Library` trait in the `definitions` project to use `NonEmptyList` instead of `List`, as well as updating the library definitions in the content repositories.

franciscodr:
This pull request adds a new functionality that allows the compiler to check if the sections list of a `Library` object is empty.

If the list is empty, the compiler will throw an error like:

This PR closes #511 

```
Unable to create SampleLibrary: A Library object must contain at least one section
```

@raulraja @dialelo Please review at your convenience. Thanks!


dialelo:
LGTM :+1:

PR #566, opened 2016-08-03
Add Docker instructions to Readme

misto:
Instead of running the database on the system directly, it can also be started inside a docker container. This pull requests adds a section about that to the Readme.


dialelo:
Thanks a lot for including Docker instructions, really appreciated!

misto:
You're welcome :-) and feel free to re-word the text, it's just a suggestion.

Issue #560, opened 2016-07-13
Running locally not showing libraries - std, cats and shapeless listing


pawank:
Hello,

"sbt run" shows below warning messages (showing tail of many such)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:35: Inferred type containing Product
[warn] GET /progress/library/:libraryName/section/:sectionName org.scalaexercises.exercises.controllers.UserProgressController.fetchUserProgressBySection(libraryName: String, sectionName: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:35: Inferred type containing Serializable
[warn] GET /progress/library/:libraryName/section/:sectionName org.scalaexercises.exercises.controllers.UserProgressController.fetchUserProgressBySection(libraryName: String, sectionName: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:35: Inferred type containing Serializable
[warn] GET /progress/library/:libraryName/section/:sectionName org.scalaexercises.exercises.controllers.UserProgressController.fetchUserProgressBySection(libraryName: String, sectionName: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:9: Inferred type containing Serializable
[warn] GET /loaderio-:token/ org.scalaexercises.exercises.controllers.LoaderIOController.verificationToken(token: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:14: Inferred type containing Serializable
[warn] GET /users/:login org.scalaexercises.exercises.controllers.UserController.byLogin(login: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:16: Inferred type containing Serializable
[warn] POST /exercises/sections/:libraryName/:sectionName/evaluate org.scalaexercises.exercises.controllers.ExercisesController.evaluate(libraryName: String, sectionName: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:20: Inferred type containing Serializable
[warn] GET /assets/*file controllers.Assets.at(path="/public", file)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:31: Inferred type containing Serializable
[warn] GET /:libraryName               org.scalaexercises.exercises.controllers.ApplicationController.library(libraryName: String)
[warn] /Users/pawan/git/github/scala-exercises/server/conf/routes:32: Inferred type containing Serializable
[...]

PR #568, opened 2016-08-09
Added troubleshooting for running locally, refering to issue #560.

diversit:
As suggested on gitter, added a section to 'Troubleshooting' about running locally. Changed the 'Troubleshooting' header so it better stands out and created sub-sections for both 'running locally' and 'ensime'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 
 ### Online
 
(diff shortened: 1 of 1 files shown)


dialelo:
Thanks a lot :clap:

PR #569, opened 2016-08-11
Added info about adding additional exercises and a Troubleshooting se…

diversit:
…ction for when exercises do not show up in the application.
As discussed on Gitter.


dialelo:
Thanks for taking the time to write it! :+1:

PR #571, opened 2016-08-11
JP - Remote Evaluator Integration

juanpedromoreno:
This PR integrates the new remote Scala evaluator within the scala-exercises project. As a part of this work, different tasks have been done:

- [x] Exercise Compiler Plugin updated in order to generate automatically, the build meta information in the different libraries (like resolvers and libraryDependencies, among others).
- [x] New Compiler Updates. Now, the compiler is able to to fetch meta data information and generate the associated code.
- [x] Evaluator Client Algebra integrated in the Scala Exercises Service
- [x] Added new Algebra and Interpreter associated to the Evaluator
- [x] The old Evaluator has been removed
- [x] There are some timeout issues between both services: scalaexercises - evaluator
- [x] Include `doobie` and  `fpinscala` libraries
- [x] Unwrap responses properly
- [x] Formatting Error Message properly
- [x] Report new issues to add unit tests that might increase the code coverage
- [x] Test everything in stage env 
- [x] Stress Testing at https://scala-exercises-pr-571.herokuapp.com

Please, @raulraja @dialelo could you review? Thanks!


compiler/src/test/scala/org/scalaexercises/exercises/compiler/CompilerSpec.scala
```diff
@@ -9,6 +9,38 @@ import scala.tools.nsc.{ Global, Settings }
...
+        /** The value is scala.collection.Seq("org.scala-lang:scala-library:2.11.7", "org.scala-lang:scala-compiler:2.11.7:ensime-internal", "org.scala-lang:scala-library:2.11.7:ensime-internal", "org.scala-lang:scala-reflect:2.11.7:ensime-internal", "org.scala-lang:scalap:2.11.7:ensime-internal", "com.chuusai:shapeless:2.2.5", "org.scalatest:scalatest:2.2.4", "org.scala-exercises:exercise-compiler:0.2.3-SNAPSHOT", "org.scala-exercises:definitions:0.2.3-SNAPSHOT", "org.scalacheck:scalacheck:1.12.5", "com.github.alexarchambault:scalacheck-shapeless_1.12:0.3.1"). */
+        val libraryDependencies: scala.collection.Seq[String] = scala.collection.Seq("org.scala-lang:scala-library:2.11.7", "org.scala-lang:scala-compiler:2.11.7:ensime-internal", "org.scala-lang:scala-library:2.11.7:ensime-internal", "org.scala-lang:scala-reflect:2.11.7:ensime-internal", "org.scala-lang:scalap:2.11.7:ensime-internal", "com.chuusai:shapeless:2.2.5", "org.scalatest:scalatest:2.2.4", "org.scala-exercises:exercise-compiler:0.2.3-SNAPSHOT", "org.scala-exercises:definitions:0.2.3-SNAPSHOT", "org.scalacheck:scalacheck:1.12.5", "com.github.alexarchambault:scalacheck-shapeless_1.12:0.3.1")
```
raulraja: @juanpedromoreno @dialelo There should be no need to provide the evaluator with a list of transitive dependencies. The evaluator should already fetch all transitive dependencies for a given maven artifact given a resolver. We should fix this as we can't expect library authors to list all their transitive dependencies here. Why is this needed in the compiler?

raulraja:
Other than my comment regarding transitive dependencies at compile time this LGTM.
@juanpedromoreno @jdesiloniz @dialelo Can we thoroughly test this to ensure appropriate errors and evaluation is shown to the end user before we merge it all?

compiler/src/test/scala/org/scalaexercises/exercises/compiler/CompilerSpec.scala
juanpedromoreno: @raulraja If library authors are using the sbt exercise compiler plugin, they won't need to add anything special, everything is figured out automatically. This example is just an spec

raulraja: What I'm saying is that it's a bug if the evaluator endpoint needs more than just the name of the artifact. For example to evaluate an exercise in the shapeless lib you just need to send the dependency for "org.scala-exercises" % "shapeless_2.11" % "1.0". The Evaluator server should automatically read the transitive dependencies and find shapeless and others that the exercise lib depends on. It's that not the case @juanpedromoreno @dialelo ?

juanpedromoreno: Sounds good @raulraja 

We could simplify it in a separate PR. Tracked issue: https://github.com/scala-exercises/scala-exercises/issues/576. Thanks.

PR #580, opened 2016-09-01
Add codecov.io support for test coverage

jdesiloniz:
This PR adds support for code test coverage through codecov.io. Could you please review, @raulraja @dialelo? Thanks a lot!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,8 +32,6 @@ lazy val formattingSettings = SbtScalariform.scalariformSettings ++ Seq(
 )
 
 lazy val scoverageSettings = Seq(
-  ScoverageKeys.coverageMinimum := 60,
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -31,7 +31,21 @@ lazy val formattingSettings = SbtScalariform.scalariformSettings ++ Seq(
       .setPreference(PlaceScaladocAsterisksBeneathSecondAsterisk, true)
 )
 
+lazy val scoverageSettings = Seq(
```
juanpedromoreno: Are these settings actually being used?

jdesiloniz: Good catch. Removing them as they're not needed (yet). It'd be better to include them when they're required.

raulraja:
LGTM

juanpedromoreno:
LGTM

Issue #520, opened 2016-07-06
Library logo images are not part of the content modules


raulraja:
The content modules via the Library trait impl should provide references to the files where the library content logo image is located. Currently it's located in the site.

jdesiloniz:
@raulraja @israelperezglez I have this mostly laid out, but I think it'd be a good idea to have a default logo for those libraries that don't come with a correct logo location. What are your thoughts about it?

israelperezglez:
@jdesiloniz I attach the icon

[scala_exercises.svg.zip](https://github.com/scala-exercises/scala-exercises/files/461111/scala_exercises.svg.zip)

PR #588, opened 2016-09-07
Load libraries logos from content libraries

jdesiloniz:
**DO NOT MERGE**: pending default logo as being discussed in #520

This PR includes code to load the content libraries logo from the libraries themselves (instead of relying on having the logo in the scala-exercises repository). This will make it easier to content providers, as only PRs from their repositories will be needed.

Fixes: #520 

Could you please review, @dialelo @raulraja @juanpedromoreno


jdesiloniz:
Just added a default logo for those libraries which don't include one

server/app/org/scalaexercises/exercises/services/ExercisesService.scala
```diff
@@ -136,11 +140,25 @@ sealed trait RuntimeSharedConversions {
...
       buildInfo = convertBuildInfo(library.buildMetaInfo)
     )
 
+  def loadLogoSrc(logoPath: String): Option[String] = {
+    def checkLogoPath(): Boolean =
+      Option(getClass.getClassLoader.getResource(logoPath + ".svg")).isDefined
+
+    def logoData(path: String): Option[String] =
+      Option(getClass().getClassLoader.getResourceAsStream(path))
+        .map(stream ⇒ new BASE64Encoder().encode(IOUtils.toByteArray(stream)))
```
juanpedromoreno: Maybe we could create a single instance of `BASE64Encoder` in this `ExercisesService` object that we could re-use each time.

juanpedromoreno:
Minor comment. Once addressed, LGTM!

PR #592, opened 2016-10-03
Change resource access from HTTP to HTTPS

### LINKED ISSUE #467: Don't include any assets that aren't served with HTTPS
Since we are serving the site through HTTPS, all the content we load from different sources than scala-exercises should be server with HTTPS too.

jdesiloniz:
Fixes #467 

This PR modifies a JS script we're using right now to fetch emoji images to add support for HTTP in order to avoid security problems. Right now it seems there are no other conflictive resources being loaded as HTTP. Could you please review, @dialelo? Thanks!! :)

### DIFF
--- a/server/public/javascripts/emojify-cdn.js
+++ b/server/public/javascripts/emojify-cdn.js
@@ -38,7 +38,7 @@
     return {
       config: {
         emojify_tag_type: 'img',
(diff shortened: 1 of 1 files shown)


dialelo:
LGTM :+1:

PR #594, opened 2016-10-04
Some cleanups (?)

peterneyens:
While I played around with the code to familiarize myself with the project, I tried to clean up some small things.
-

### DIFF
--- a/runtime/src/main/scala/org/scalaexercises/exercises/Exercises.scala
+++ b/runtime/src/main/scala/org/scalaexercises/exercises/Exercises.scala
@@ -13,6 +13,8 @@ import java.net.URLClassLoader
 import java.io.File
 
 import cats.data.Xor
+import cats.std.list._
+import cats.syntax.monadCombine._
 import org.clapper.classutil.ClassFinder
 import org.scalaexercises.evaluator.Dependency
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/scala/org/scalaexercises/exercises/Exercises.scala
```diff
@@ -40,24 +42,18 @@ object Exercises {
...
-
-      // until a bifoldable exists in Cats...
-      loadedLibrary match {
-        case Xor.Right(c) ⇒ (acc._1, c :: acc._2)
-        case Xor.Left(e)  ⇒ (e :: acc._1, acc._2)
-      }
     }
 
-    (errors, libraries)
+    errorsAndLibraries.separate
```
peterneyens: Cats has `Bifoldable` now, but we can use `MonadCombine.separate` (which uses `Bifoldable` underneath) here to split the list of errors and libraries in two lists.

server/app/org/scalaexercises/exercises/services/ExercisesService.scala
```diff
@@ -26,12 +28,9 @@ object ExercisesService extends RuntimeSharedConversions {
...
 
   lazy val (libraries: List[Library], librarySections: Map[String, List[Section]]) = {
-    val libraries1 = colorize(runtimeLibraries)
+    val libraries1 = colorize(runtimeLibraries).map(convertLibrary)
     errors.foreach(error ⇒ Logger.warn(s"$error")) // TODO: handle errors better?
-    (
-      libraries1.map(convertLibrary),
-      libraries1.map(library0 ⇒ library0.name → library0.sections.map(convertSection)).toMap
-    )
+    (libraries1, libraries1.map(lib ⇒ lib.name → lib.sections).toMap)
```
peterneyens: I think `convertSection` was called twice for every section, once in `convertLibrary` and once explicitly when creating the `Section` map).

server/app/org/scalaexercises/exercises/services/ExercisesService.scala
```diff
@@ -110,29 +105,31 @@ sealed trait RuntimeSharedConversions {
...
-          buildMetaInfo = library.buildMetaInfo
-        ) :: librariesAcc)
-      } else
-        colors → (library :: librariesAcc)
-    }
-    res.reverse
+      } else {
+        State.pure[Colors, RuntimeLibrary](library)
+      }
+    }.runA(autoPalette).value
```
peterneyens: I prefer the `State[Colors, ?]` with `traverse` over the `foldLeft`, but opinions may vary ?

dialelo:
Awesome, LGTM :+1:

PR #595, opened 2016-10-05
Add the canonical url to every page.

### LINKED ISSUE #452: Include canonical URLs in the page's head
Since we are redirecting users to the HTTPS and www-prefixed version of the site, we should add the [canonical URL](https://moz.com/blog/canonical-url-tag-the-most-important-advancement-in-seo-practices-since-sitemaps) to every page so we don't get any SEO penalization for "duplicated" content.

peterneyens:
Fixes #452.

### DIFF
--- a/server/app/views/templates/base/head.scala.html
+++ b/server/app/views/templates/base/head.scala.html
@@ -2,23 +2,25 @@
 @import org.scalaexercises.exercises.utils.StringUtils._
 @(title: String, description: String )(implicit request: RequestHeader)
 
+@baseUrl = @{ current.configuration.getString("application.url").getOrElse("https://www.scala-exercises.org") }
 
 <meta charset="UTF-8">
 <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <link rel="shortcut icon" type="image/png" href="@routes.Assets.at("images/favicon.png")" />
+<link rel="canonical" href="@baseUrl@request.uri" />
 
 <title>@title.humanizeUnderscore</title>
 <meta name="description" content="@description.humanizeUnderscore">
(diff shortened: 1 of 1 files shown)


server/app/views/templates/base/head.scala.html
```diff
@@ -2,23 +2,25 @@
...
 <meta name="description" content="@description.humanizeUnderscore">
 <meta name="author" content="47 Degrees">
-<meta property="og:image" content="@current.configuration.getString("application.url").getOrElse("https://www.scala-exercises.org")@routes.Assets.at("images/poster.png")"/>
+<meta property="og:image" content="@baseUrl@routes.Assets.at("images/poster.png")"/>
 <meta property="og:title" content="@title.humanizeUnderscore"/>
 <meta property="og:site_name" content="@current.configuration.getString("application.name").getOrElse("Scala Exercises")"/>
-<meta property="og:url" content="@current.configuration.getString("application.url").getOrElse("https://www.scala-exercises.org")@request.path"/>
-<meta property="og:type" content="website" />
+<meta property="og:url" content="@baseUrl@request.path"/>
+<meta property="og:type" content="website" />s
```
dialelo: That rogue `s`, let's :fire: it and merge this!

raulraja:
@peterneyens I need to understand what '@baseUrl' does to get it's information. If it's loading from the config is fine because we are setting a value for that there but if it's using `request.host` it's not gonna work on heroku because we are behind a load balancer there and we've had issues with that before in similar scenarios. https://github.com/scala-exercises/scala-exercises/pull/482/files

peterneyens:
@raulraja `@baseUrl` is defined in the same file, I just created this alias because there were already three other places where `application.url` was loaded from the config file.

raulraja:
Makes sense! I missed that and thought it was a built in function.

raulraja:
Thanks @peterneyens !

PR #596, opened 2016-10-05
Remove the trailing slash from a route if no handler is found.

### LINKED ISSUE #563: Trailing / in url at library root fails with action not found
If you navigate to the root url for a library and include a trailing / then it fails to find the route. See for example:

https://www.scala-exercises.org/shapeless/ vs https://www.scala-exercises.org/shapeless
https://www.scala-exercises.org/std_lib/ vs https://www.scala-exercises.org/std_lib

screen shot 2016-07-27 at 1 26 38 pm

peterneyens:
Based on the code used in the Play Framework website.

Fixes #563.


dialelo:
LGTM :+1:

PR #602, opened 2016-10-19
Refactor methods in UserProgressOps

peterneyens:
I created a few helper methods and I replaced
- `FM.sequence( xs.map { ... } )`  with `xs.traverse`,
- some for comprehensions with a simple `map` and
- the pattern `maybeLibrary.fold(Nil: List[Sections])(_.sections)` with `maybeLibrary.foldMap(_.sections)`.


raulraja:
This looks great @peterneyens. Can we add some tests for this?

PR #607, opened 2016-11-08
Fixing problems with input fields resize on paste events

### LINKED ISSUE #606: Pasting answers does not trigger field resize and Run button
**Description:**
  When pasting data into an answer checkbox, the field resize does not happen and the Run button does not activate. 

**Repro Steps:**
   1)  Pick a question, any question
   2)  Select text from the problem above "We just encountered an awesome Woodchuck Stampede!" and Copy
   3)  Paste the text into the answer box

**Observed Behavior:**
  1)  The answer field does not resize to show the entire text
  2)  The Run button does not enable

**Expected Behavior:**
  1)  The answer field should resize to show the entire text
  2)  The Run button should enable

**Notes:**
  The event that triggers these two changes appears to be based on keystrokes that change the contents of the field, because changing focus with the mouse or Tab key does not cause an update.

If possible the updates should be triggered by a text/contentChanged event, so that any changes to the contents of the field would trigger the updates.

jdesiloniz:
Fixes #606 

This PR includes code to fix paste events not being correctly handled by the input fields in the exercises. It also replaces the resizing code, as the resulting size wasn't accurate for long texts.

Could you please review, @dialelo @rafaparadela @juanpedromoreno? Thanks!

### DIFF
--- a/client/src/main/scala/utils/DomHandler.scala
+++ b/client/src/main/scala/utils/DomHandler.scala
@@ -11,6 +11,7 @@ import scala.scalajs.js
 import org.scalajs.dom
 import org.scalajs.dom.raw.{ HTMLDivElement, HTMLElement, HTMLInputElement }
 import org.scalajs.jquery.{ jQuery ⇒ $, JQuery }
+import scala.scalajs.js.timers._
 
 import monix.cats._
 import monix.eval.Coeval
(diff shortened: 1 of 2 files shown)


client/src/main/scala/utils/DomHandler.scala
```diff
@@ -70,31 +71,85 @@ object DomHandler {
...
   } yield ()
 
+  /** Assigns behavior to an input element change event
+    */
+  def onInputChange(
+    onchange: (String, Seq[String]) ⇒ Coeval[Unit]
+  ): Coeval[Unit] = {
+    for {
+      inputs ← allInputs
+      _ ← inputs.map(input ⇒ attachOnInputChange(input, onchange)).sequence
```
raulraja: This can be done in a single iteration 
```scala
inputs.map { input => attachOnInputChange(input); attachOnInputPaste(input) }.sequence
```

client/src/main/scala/utils/DomHandler.scala
```diff
@@ -70,31 +71,85 @@ object DomHandler {
...
+
   def attachKeyUpHandler(
     input:          HTMLInputElement,
     onkeyup:        (String, Seq[String]) ⇒ Coeval[Unit],
     onEnterPressed: String ⇒ Coeval[Unit]
   ): Coeval[Unit] = Coeval {
     $(input).keyup((e: dom.KeyboardEvent) ⇒ {
+      updateExerciseInput(
+        input,
+        (info: (String, Seq[String])) ⇒ {
```
raulraja: Let's destructure this tuple in the match below and also give it some good identifiers instead of `._1` `._2`

Issue #616, opened 2017-01-23
Add Scala Mooc exercises to the platform


anamariamv:
We should add Scala MOOC exercises to the platform

juanpedromoreno:
Done in https://github.com/scala-exercises/scala-exercises/pull/618

PR #617, opened 2017-01-24
Adds scala-tutorial Library

### LINKED ISSUE #603: Scala Exercises based off of MOOC curriculum
Hi guys,

I'm not sure if this is the right place to be posting this as this is a content-related point rather than a technical one. So feel free to close/move it if it should be somewhere else.

But in continuing to work on the Scala MOOCs, I just realized that I would love to point people to scala-exercises.com from the MOOC videos and resource pages. In particular for the Spark course, because we expect that people will skip over the previous 3 MOOCs introducing Scala + the fundamentals, just being in a rush to learn Spark.

So I started looking around for something interactive that I could point people at. I immediately thought of scala-exercises, but realized that the stdlib exercises is maybe not the best place to point these people. Not that there's anything wrong with it, it's just maybe not the best beginner-first place to send people trying to get a quick idea of what Scala is all about. 

That said, I was thinking it would be cool if there was a set of exercises based on the first 2 MOOCs (Functional Programming Principles in Scala and Functional Program Design in Scala). The progression that Martin has designed for these courses is really spot on, if you ask me (of course I'm biased). 

Just brainstorming here. Would something like this ever be feasible? (Again, just trying to figure out what I should point fast-track students at)

juanpedromoreno:
**DO NOT MERGE YET**

Fixes #603 and #616

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,12 @@ env:
   global:
     JAVA_OPTS=-Xmx2g
     SBT_OPTS="-XX:+UseConcMarkSweepGC -XX:MaxPermSize=512m"
+
 before_install:
-  - openssl aes-256-cbc -K $encrypted_ef7b2581dd2c_key -iv $encrypted_ef7b2581dd2c_iv -in secring.gpg.enc -out secring.gpg -d
+- if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
+     openssl aes-256-cbc -K $encrypted_ef7b2581dd2c_key -iv $encrypted_ef7b2581dd2c_iv -in secring.gpg.enc -out secring.gpg -d;
+  fi
+
 install:
   - gem install sass
 script:
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ $ docker run --name scala-exercises-db -e POSTGRES_DB=scalaexercises_dev -e POST
 
 ##### Configure the application
 
-Edit the `site/server/conf/application.dev.conf` configuration file with your database information.
+Edit the `server/conf/application.dev.conf` configuration file with your database information.
 
 #### Running the app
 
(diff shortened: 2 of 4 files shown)


juanpedromoreno:
Hi @julienrf, I've deployed in this stage env the new library. Everything looks good, however, I can see how in the first chapter, in the second exercise, the expected types are interchanged:

https://scala-exercises-pr-617.herokuapp.com/scala_tutorial/terms_and_types

Please, correct me if I'm missing something. Thanks.

julienrf:
Oh, good catch! Indeed, it seems that I can not use `res1` *before* using `res0`, here: https://github.com/scala-exercises/exercises-scalatutorial/blob/master/src/main/scala/scalatutorial/sections/TermsAndTypes.scala#L123-L124

It might be an issue of scala-exercise?

Anyway, I’ll fix the exercise definition by swapping the arguments order.

julienrf:
So, I went through all the other exercises and they are fine, excepted that we can not run an exercise that takes no parameters. I have an exercise that does not take parameters (ie. where users have nothing to fill in) in order to show the output of the compilation. Also, sometimes I don’t end a section with an exercise, so I put a dummy function instead, but this function is not runnable from the UI.

Do you think it would be possible to make these no-args functions runnable from the UI?

julienrf:
BTW, I just pushed [this commit](https://github.com/scala-exercises/exercises-scalatutorial/commit/fed99499f32737d1327ff8b06ad405210ff6e4dc) to the scalatutorial repo.

juanpedromoreno:
It might be possible if we modify the UI, activating the `Run` button for those exercises that contains code but don't have any exercise.

julienrf:
@juanpedromoreno I’m not sure it will be enough: these disabled exercises must also not be taken into account by the progress bar:

progress

juanpedromoreno:
Right @julienrf That's an issue.

juanpedromoreno:
Closing and creating a new PR in order to re-deploy Heroku.

https://github.com/scala-exercises/scala-exercises/pull/618

juanpedromoreno:
I've submitted the issue: https://github.com/scala-exercises/scala-exercises/issues/619

PR #618, opened 2017-01-25
Adds scala-tutorial Library - RE

### LINKED ISSUE #603: Scala Exercises based off of MOOC curriculum

juanpedromoreno:
Fixes #603 and #616 

Well-known issue that we should address in a separate PR: https://github.com/scala-exercises/scala-exercises/issues/619

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,6 +51,7 @@ lazy val server = (project in file("server"))
       "org.scala-exercises" %% "exercises-doobie" % version.value changing(),
       "org.scala-exercises" %% "exercises-scalacheck" % version.value changing(),
       "org.scala-exercises" %% "exercises-fpinscala" % version.value changing(),
+      "org.scala-exercises" %% "exercises-scalatutorial" % version.value changing(),
       "org.scala-exercises" %% "runtime" % version.value changing(),
       "org.scala-exercises" %% "evaluator-client" % v('evaluator) changing(),
       "org.slf4j" % "slf4j-nop" % v('slf4j),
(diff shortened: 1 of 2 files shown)


juanpedromoreno:
@julienrf Do you think https://scala-exercises-pr-618.herokuapp.com/ is now ready to go?

Thanks!

julienrf:
Yes, thanks!

juanpedromoreno:
@raulraja We have the same issue in master branch, I've just relaunched the last build getting the same issue https://travis-ci.org/scala-exercises/scala-exercises/builds/186760797 . 

I'd suggest addressing this problem in a separate PR.

PR #651, opened 2017-05-23
Add debug error handler

raulraja:
Attempt to troubleshoot prod issues that we can't reproduce locally.
In prod after a few mins the index seems to through an uncaught exception that we can hopefully see in the logs once this is deployed.


FPerezP:
Looks good!

Issue #663, opened 2017-07-26
Still not able to run Exhaustivity (Scala Structuring Information)


andys8:
Exhaustivity can't be run. Should be possible with #619 but I can't confirm it and there are more people having trouble (disqus, github issue).

https://www.scala-exercises.org/scala_tutorial/structuring_information

![img_0092](https://user-images.githubusercontent.com/13085980/28633964-ddfb4b40-7235-11e7-9806-ff5816e18feb.PNG)

andys8:
[Here](https://www.scala-exercises.org/scala_tutorial/classes_vs_case_classes) is another one. The last block contains only `()` and can't be executed.

AdrianRaFo:
Maybe we need to update the version and publish a new release because the last exercise is in the code, the `lazy evaluation` was fixed and still with the same error

andys8:
@AdrianRaFo Do you know who will do the release?

andys8:
[And another one](https://www.scala-exercises.org/scala_tutorial/lazy_evaluation)
<img width="1084" alt="screen shot 2017-08-05 at 16 22 43" src="https://user-images.githubusercontent.com/13085980/28996151-5bf050ce-79fa-11e7-8f7b-ca5112e80bfb.png">

AdrianRaFo:
Sorry, I don't know but it has other problems so maybe we should wait until be fixed to launch a new release

raulraja:
Was the version ever updated? I can do a new release that gets the latest snapshots if the exercise was fixed. Is there a PR where this was fixed @AdrianRaFo?

AdrianRaFo:
I launched a pr that fixed a error in one of the exercise but I think that the new release wasn't launched when the exercises were fixed

raulraja:
Is it included here: https://github.com/scala-exercises/scala-exercises/pull/665 ?

AdrianRaFo:
That pr update the library to `0.4.2-SNAPSHOT` but the next release on scala tutorial should be `0.4.1-SNAPSHOT`

raulraja:
@dominv can you include these changes it in the next master build? thanks

juanpedromoreno:
`0.4.2-SNAPSHOT` supersedes to `0.4.1-SNAPSHOT`, the current code at https://github.com/scala-exercises/scala-exercises/pull/665 will release our latest changes across all libraries. Hence, the next deployment will upgrade the entire platform.

dominv:
[...]

PR #665, opened 2017-08-04
Updating libraries versions

dominv:
This PR updates the scala versions to 2.11.11 and 2.12.3 , sbt-org-policies to 0.5.12 and all the scala-exercises libraries to their last version

@juanpedromoreno could you please review?

Thanks!


raulraja:
@dominv how far are we from getting this ready?, seems like merging it would fix some exercises that are currently broken in the live version. See #663

PR #667, opened 2017-08-30
Fix exercises consuming 20% CPU due to excessive CSS animation

ronjouch:
I thought I was in browser/extension trouble, but getting feedback from
https://github.com/webcompat/web-bugs/issues/9410 revealed the problem
happens in both Chrome and Firefox.

Looks like all these animations are pretty CPU intensive. Maybe there's
a better solution to this, so consider this PR a glorified issue, and feel free
to rework/change if you find a proper solution rather than my disabling.

### DIFF
--- a/server/app/assets/stylesheets/layout/_exercises.scss
+++ b/server/app/assets/stylesheets/layout/_exercises.scss
@@ -60,7 +60,9 @@
         input {
             color: $input-unsolved;
             border-bottom: 1px solid $input-unsolved;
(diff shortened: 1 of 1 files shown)


calvellido:
Hello @ronjouch, thank you very much for your contribution. We are trying to solve the problem without changing the input's current behaviour on #670. Only highlighting the input on hover or focus could cause problems for a user to notice on where to insert the solution to each exercise.

I will try to explain the proposed solution on #670 too.

Please have a look at it and thanks for noticing and proposing a solution!

PR #670, opened 2017-09-25
Change exercises input animation method

calvellido:
* Use one step animation, instead of a linear one, so we avoid repainting continuously
* Add will-change property as is the actual way on informing the browser that an element will not stay the same
* Add also null 3D transform to lay the CPU burden down, as IE or Edge doesn't have will-change implemented as of yet

### DIFF
--- a/server/app/assets/stylesheets/layout/_exercises.scss
+++ b/server/app/assets/stylesheets/layout/_exercises.scss
@@ -25,7 +25,7 @@
         padding: 30px;
         box-shadow: inset 3px 0 rgba(0,0,0, 0.1);
     }
-    
+
     .solved-exercise {
         border: 2px solid #9CCC65;
         box-shadow: none;
@@ -44,12 +44,12 @@
                 button {
                     transition: opacity 0.5s linear;
                 }
-                 
+
                 i {
                     margin-right: 4px;
                     font-size: 10px;
                 }
-                
+
                 button:hover,
                 button:focus,
                 button:active {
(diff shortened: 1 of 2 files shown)


calvellido:
What causes more _pain_ to a browser performance is [repainting areas](https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas) as the browser needs to do some expensive calculations of a lot of elements in a web layout. There's a lot of things that can trigger a repaint, being obvious an animation set on an element.

Due to the fact that this animation added to the user input on Scala Exercises is being executed on an infinite basis, the CPU load is gonna take a hit by this.

Also, there are some operations which a browser can really animate cheaply (basically position translations, basic scaling and rotation, and opacity), being the border styling, which is the one included on the Scala Exercises user input boxes, a _not-so-cheap_ one.

The proposed changes tackled these problem in various ways.

* First of them, as our animation is not really in a need of smoothness or full transition as it is a mimic of terminals (or any other device or software) user input, we can set the `animation-timing-function` to one that only uses one step on it, thus triggering only a repaint per animation time completion.

* Also we are adding the `will-change` property to inform properly to the browser that it is gonna need to take care of that elements performance in advance, as it is gonna change over time.

* Besides, [though in consideration, the will-change property is not yet implemented in Microsoft Edge browsers](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/csswillchange/), so also the _null transform hack_ is added to force the browser to set that element in a new layer and making the GPU to take care of it when available.

I'm adding a comparison between current performance load (first screen) and the combination of settings and properties of this PR (second screen):

not-using-steps
using_steps

For a more dramatic and real-world comparison the CPU is throttled by 5x. Also, to isolate the animation load hit, the animations were paused in intervals of 5 seconds of a full 30s profiling (which can be seen see on the first profiling screenshot) of the Scala Exercises STD Lib Asserts section, which includes three of these kind of inputs.

PR #672, opened 2017-10-08
Better handle syntax sugar for compose examples

ml10:
When a block starts with a value that uses the `(a b c)` syntax sugar for `a.b(c)`, the leading `(` is stripped.

Two examples of this are:

- [Applicative composition](https://www.scala-exercises.org/cats/applicative) in the cats exercises
```scala
Applicative[List] compose Applicative[Option]).pure(1) ...
```
- [Lens composition](https://www.scala-exercises.org/monocle/lens) in the monocle exercises.
```scala
addressLens composeLens streetNumber).get(john) ...
```
I wanted to keep the unit test simple and not worry about imports, so I used composition of the List.apply method for indexing.

I struggled a bit to figure out how to version everything with the plugins in place to be able to test this running scala-exercises locally. Ivy wasn't pulling in my locally published snapshots and rather than futz too much with `sbt-dirty-money` or something, I thought I'd get a PR out for comments first.

### DIFF
--- a/compiler/src/main/scala/org/scalaexercises/exercises/compiler/MethodBodyReader.scala
+++ b/compiler/src/main/scala/org/scalaexercises/exercises/compiler/MethodBodyReader.scala
@@ -62,7 +62,10 @@ object MethodBodyReader {
         )
         (start0, end)
       case _ ⇒
-        (tree.pos.start, tree.pos.end)
+        val source      = tree.pos.source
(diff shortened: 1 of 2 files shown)


ml10:
I should note the code changes for this were inspired by `scala.reflect.internal.util.InternalPositionImpl#calculateColumn`

PR #676, opened 2018-01-10
Validating user progress regarding exercises subject to change

jdesiloniz:
Fixes: https://github.com/scala-exercises/exercises-cats/issues/65

This PR tries to fix an issue related to how we handle user progress in the application. In some circumstances an user can have progress stored for an exercise and when coming back to it later this exercise can be changed. If the exercise has less fields than before (i.e.: it's been splitted into two), these extra stored values will still be sent to the external compiler and any evaluation will be rejected for carrying extra parameters for the exercise function.

What this code does is simply verifying the state just after retrieval, to check if we're storing more values than those actually required, and if that's the case invalidating the progress for that exercise in particular (which in my opinion should be desirable, as the exercise has been modified anyway).

### DIFF
--- a/client/src/main/scala/Effects.scala
+++ b/client/src/main/scala/Effects.scala
@@ -44,7 +44,9 @@ object Effects {
       Client
         .fetchProgress(lib, sect)
         .collect({
-          case Some(state) ⇒ Some(SetState(state))
+          case Some(state) ⇒ {
+            Some(SetState(validateState(state)))
+          }
         })
     })
   }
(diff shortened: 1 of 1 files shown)


client/src/main/scala/Effects.scala
```diff
@@ -63,4 +65,19 @@ object Effects {
...
     }
 
+  def validateState(state: List[ClientExercise]): List[ClientExercise] =
+    state.foldLeft(List.empty[ClientExercise]) {
+      (acc: List[ClientExercise], exercise: ClientExercise) =>
+        DomHandler
+          .findExerciseByMethod(exercise.method)
+          .map(DomHandler.inputsInExercise)
+          .fold(acc) { inputs =>
+            if (exercise.arguments.size <= inputs.size) {
```
teliatko: Is it not better to always invalidate the exercise, I mean when:
`if (exercise.arguments.size == inputs.size) {
acc :+ exercise
} else {
acc :+ exercise.copy(arguments = Seq.empty, state = Unsolved)
}
`
I understand that you cannot better detect if the exercise was changed or not, but if it was the change can be anything, even the change of the params order, thus is better to mark it as Unresolved.

I dunno how `Unsolved` is interpreted in UI.

IMO exercise should be correctly set as `Unsloved` when there is a change. Then you can choose if the arguments will be preserved, always, i.e. striped when they are too long or always cleared.

teliatko: Otherwise it solves my issue.

jdesiloniz: Yeah the downside of this is that we're storing in the DB the user progress, which basically carries the following data: which exercise was, which fields were completed, and what was the state of the exercise in that moment (completed successfully or not).

In any case you're right. I'm going to extend this fix to all exercises that have changed the number of arguments from the user progress (even if they've added stuff). We don't have a way to know if the order or expected values changed, but I think it's sensible. I still think that invalidating the fields is the safest option here, but maybe other members of the team can input some suggestions about this.

Thanks!

PR #678, opened 2018-02-04
Expand error log area to fill all content

pvorona:
This change will make block with error log expand to fill whole message.
Before:
screen shot 2018-02-04 at 02 55 40
After:
screen shot 2018-02-04 at 02 55 53


calvellido:
Thank you very much for your contribution @pvorona 🤙

PR #687, opened 2019-06-21
Updating JDK version for travis

jdesiloniz:
Updating to OpenJDK, to avoid broken compilations in PRs reviews, etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   - . $HOME/.nvm/nvm.sh
   - nvm install stable
   - nvm use stable
-  - npm install jsdom
+  - npm install jsdom@9.12.0
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ before_script:
 scala:
   - 2.11.11
 jdk:
-  - oraclejdk8
+  - openjdk11
```
JesusMtnez: I think you could use `opendjk8` if you prefer to use the same version

jdesiloniz: @JesusMtnez Thanks. I'm wary of creating more issues. We should consider updating the Scala/JDK versions but that would imply reviewing all exercises and check nothing is breaking.

PR #692, opened 2019-09-25
Dependencies and settings update

kiroco12:
This PR updates the dependencies to their latest (and compatible) versions. These are not the final dependencies, so in a future the list can be changed.


build.sbt
```diff
@@ -49,30 +51,30 @@ lazy val server = (project in file("server"))
...
-      "com.vmunier"             %% "scalajs-scripts"  % v('scalajsscripts),
-      "com.lihaoyi"             %% "upickle"          % v('upickle),
-      "org.webjars"             %% "webjars-play"     % v('webjars),
-      "org.webjars"             % "bootstrap-sass"    % v('bootstrap),
-      "org.webjars"             % "highlightjs"       % v('highlightjs),
-      "com.tristanhunt"         %% "knockoff"         % v('knockoff),
-      "com.newrelic.agent.java" % "newrelic-agent"    % v('newrelic),
-      "commons-io"              % "commons-io"        % v('commonsio),
-      %%("freestyle"),
+      //"org.scala-exercises" %% "exercises-stdlib" % v('stdlib) xscalaExercises,
```
calvellido: Could we simply remove the unnecessary dependencies, even it they could come back in the future.

project/plugins.sbt
```diff
@@ -4,21 +4,20 @@ logLevel := Level.Warn
...
+addSbtPlugin("com.eed3si9n"       % "sbt-buildinfo"            % "0.9.0")
+addSbtPlugin("com.47deg"          % "sbt-org-policies"         % "0.12.0-M3")
+addSbtPlugin("com.typesafe.play"  % "sbt-plugin"               % "2.8.0-M5")
+addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "0.6.1")
+addSbtPlugin("com.vmunier"        % "sbt-web-scalajs"          % "1.0.9-0.6")
+addSbtPlugin("com.typesafe.sbt"   % "sbt-gzip"                 % "1.0.2")
 
 // Scripted
-libraryDependencies += "org.scala-sbt" % "scripted-plugin" % sbtVersion.value
+libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value
```
calvellido: Is this ok? :thinking:

kiroco12: Yes, needs the double '%' to work

client/src/main/scala/scripts/ExercisesJS.scala
```diff
@@ -1,7 +1,5 @@
 /*
- *  scala-exercises
- *
- *  Copyright 2015-2017 47 Degrees, LLC. <http://www.47deg.com>
+ * Copyright 2016-2019 47 Degrees, LLC. <http://www.47deg.com>
```
vejeta: It is 2015 everywhere, why change it to 2016 here?

client/src/main/scala/utils/DomHandler.scala
```diff
@@ -1,7 +1,5 @@
 /*
- *  scala-exercises
- *
- *  Copyright 2015-2017 47 Degrees, LLC. <http://www.47deg.com>
+ * Copyright 2016-2019 47 Degrees, LLC. <http://www.47deg.com>
```
vejeta: It is 2015 everywhere, why change it to 2016 here?

client/src/main/scala/scripts/ExercisesJS.scala
kiroco12: In build.sbt in the copyright part says 2015 but the `startDate` option says 2016. Which one is correct?

build.sbt
```diff
@@ -137,43 +124,42 @@ lazy val definitions = (project in file("definitions"))
...
-      "org.scalariform" %% "scalariform"   % v('scalariform),
-      "org.scala-lang"  % "scala-compiler" % scalaVersion.value % "compile",
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value % "compile",
       %%("github4s"),
       %%("cats-core") % "compile",
       %%("cats-laws") % "test",
       %%("scalatest") % "test"
-    )
+    ),
+    addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.8.3")
```
AdrianRaFo: TODO move to scalafmt

project/ProjectPlugin.scala
```diff
@@ -83,37 +87,38 @@ object ProjectPlugin extends AutoPlugin {
         organizationEmail = "hello@47deg.com"
       ),
       orgLicenseSetting := ApacheLicense,
-      scalaVersion := "2.11.11",
+      scalaVersion := "2.12.10",
       scalaOrganization := "org.scala-lang",
-      crossScalaVersions := Seq("2.11.8"),
+      crossScalaVersions := Seq("2.12.10"),
```
AdrianRaFo: if we are going to just maintain the scala 2.12 version you don't need the crossCompile setting

vejeta:
@kiroco12 Are the errors in Travis expected?

kiroco12:
@vejeta yes, are expected. There isn't any code modifications so it won't compile

build.sbt
```diff
@@ -191,23 +177,23 @@ lazy val `sbt-exercise` = (project in file("sbt-exercise"))
...
-  .enablePlugins(BuildInfoPlugin)
+// scripted plugin
+.enablePlugins(SbtPlugin)
+.settings(
+  scriptedLaunchOpts := {
+    scriptedLaunchOpts.value ++
+      Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
+  },
+  scriptedBufferLog := false,
+//  // Publish definitions before running scripted
```
juanpedromoreno:
```suggestion
  // Publish definitions before running scripted
```

build.sbt
```diff
@@ -137,43 +124,42 @@ lazy val definitions = (project in file("definitions"))
...
+    ),
+    addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.8.3")
   )
   .dependsOn(definitions, runtime)
 
 // Compiler plugin
 
 lazy val `sbt-exercise` = (project in file("sbt-exercise"))
-  .disablePlugins(CoursierPlugin)
+//
```
juanpedromoreno:
```suggestion

```

build.sbt
```diff
@@ -16,23 +18,21 @@ lazy val `scala-exercises` = (project in file("."))
 
 // Purely functional core
 
-lazy val core = (crossProject in file("core"))
+lazy val core = (crossProject(JSPlatform, JVMPlatform) in file("core"))
```
AdrianRaFo: I think we just need JVMPlatform, aren't we?

kiroco12: The core project needs cross compilation because the client depends on the js core and the server needs the jvm core.

PR #693, opened 2019-09-27
Core module update

kiroco12:
This PR converts the old Free algebra to a Tagless-Final algebra. In consequence, some libraries have been removed from build.sbt. The PR has no changes in other modules so the rest maybe won't compile.

### DIFF
--- a/core/shared/src/main/scala/org/scalaexercises/algebra/progress/exerciseProgress.scala
+++ b/core/shared/src/main/scala/org/scalaexercises/algebra/progress/exerciseProgress.scala
@@ -70,7 +70,7 @@ class UserExercisesProgress[F[_]: Monad](implicit UO: UserProgressOps[F], EO: Ex
     user.fold(anonymousUserProgressByLibrary(libraryName))(
       fetchUserProgressByLibrary(_, libraryName))
 
-  private[this] def anonymousUserProgressByLibrary(libraryName: String): F[LibraryProgress] = {
+  private[this] def anonymousUserProgressByLibrary(libraryName: String): F[LibraryProgress] =
     for {
       lib ← EO.getLibrary(libraryName)
       sections = lib.foldMap(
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/org/scalaexercises/algebra/progress/exerciseProgress.scala
```diff
@@ -140,10 +104,9 @@ class UserExercisesProgress[F[_]](implicit UO: UserProgressOps[F], EO: ExerciseO
   }
 
   def fetchUserProgressByLibrarySection(
-                                         user: User,
-                                         libraryName: String,
-                                         sectionName: String
-                                       ): FreeS[F, SectionExercises] = {
+      user: User,
+      libraryName: String,
+      sectionName: String): F[SectionExercises] = {
```
AdrianRaFo: `{}` :fire: ?

core/shared/src/main/scala/org/scalaexercises/algebra/progress/progress.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def isSectionCompleted(user: User, libraryName: String, section: Section): F[Boolean]
+
+}
+
+abstract class UserProgressOps[F[_]: Functor] extends UserProgressOpsAlgebra[F] {
+  override def getSolvedExerciseCount(user: User, library: String, section: String): F[Int] =
+    getExerciseEvaluations(user, library, section).map(tried ⇒ tried.count(_.succeeded))
+
+  override def isSectionCompleted(user: User, libraryName: String, section: Section): F[Boolean] =
```
AdrianRaFo: an implementation of a not defined method doesn't need the override

purrgrammer:
Looks good, great job @kiroco12 :sparkles:  

It seems that the tests in Travis are failing, we need to make sure they are green before merging :green_heart:

PR #717, opened 2020-01-09
Updating to Scala 2.13.1 and bug fixes

kiroco12:
This PR updates part of the modules to Scala 2.13.1, but keeps mantaining the 2.12.10 version for compatibility with sbt-buildinfo. Also, fixes a problem with the number of completed exercises, for the people which made older exercises what now are deleted. Still have to fix a deprecation with the old collection API of Scala 2.12 in the compiler module.


build.sbt
```diff
@@ -103,14 +105,15 @@ lazy val client = (project in file("client"))
...
+    scalacOptions += "-Ymacro-annotations",
     libraryDependencies ++= Seq(
-      "io.monix"      %%% "monix"     % v('monix),
-      "org.typelevel" %%% "cats-core" % v('catsversion),
-      "com.lihaoyi" %%% "scalatags" % v('scalatags) xscalajs,
-      "org.scala-js" %%% "scalajs-dom" % v('scalajsdom),
-      "be.doeraene" %%% "scalajs-jquery" % v('scalajsjquery) xscalajs,
-      "com.lihaoyi" %%% "upickle" % v('upickle),
-      "com.lihaoyi" %%% "utest"   % v('utest) % Test
+      "io.monix"      %%% "monix"     % v(Symbol("monix")),
```
juanpedromoreno: Let's replace the symbol usage by standard strings

build.sbt
```diff
@@ -119,12 +122,13 @@ lazy val `evaluator-client` = (project in file("eval-client"))
...
-      "io.circe"      %% "circe-generic"       % v('circeversion),
-      "org.scalatest" %% "scalatest"           % v('scalatest) % Test
-    )
+      "org.http4s"    %% "http4s-blaze-client" % v("http4s"),
+      "org.http4s"    %% "http4s-circe"        % v("http4s"),
+      "io.circe"      %% "circe-core"          % v("circeversion"),
+      "io.circe"      %% "circe-generic"       % v("circeversion"),
+      "org.scalatest" %% "scalatest"           % v("scalatest") % Test
+    ),
+    crossScalaVersions := Seq(scala_212, scala_213)
```
juanpedromoreno: We don't need to cross-compile this project, having Scala 2.13 would be enough

build.sbt
```diff
@@ -135,11 +139,12 @@ lazy val definitions = (project in file("definitions"))
...
-      "org.scalatest"              %% "scalatest"                 % v('scalatest),
-      "org.scalacheck"             %% "scalacheck"                % v('scalacheckversion),
-      "com.github.alexarchambault" %% "scalacheck-shapeless_1.14" % v('scalacheckshapeless)
-    )
+      "org.typelevel"              %% "cats-core"                 % v("catsversion"),
+      "org.scalatest"              %% "scalatest"                 % v("scalatest"),
+      "org.scalacheck"             %% "scalacheck"                % v("scalacheckversion"),
+      "com.github.alexarchambault" %% "scalacheck-shapeless_1.14" % v("scalacheckshapeless")
+    ),
+    crossScalaVersions := Seq(scala_212, scala_213)
```
juanpedromoreno: Same here

build.sbt
```diff
@@ -177,11 +191,13 @@ lazy val compiler = (project in file("compiler"))
 lazy val `sbt-exercise` = (project in file("sbt-exercise"))
   .settings(name := "sbt-exercise")
   .settings(
-    scalaVersion := "2.12.10",
+    scalaVersion := scala_212,
```
juanpedromoreno: Shouldn't be Scala 2.13?

kiroco12: There's a dependency (`sbt-buildinfo`) which isn't updated to Scala 2.13 yet, so the plugin can't be updated to Scala 2.12.

build.sbt
kiroco12: Libraries with dependencies used in the `sbt-exercise` project need a version in Scala 2.12. That's why cross-compilation.

PR #846, opened 2020-08-26
Downgrade pureconfig-exercises version

paualarco:


BenFradet:
Indeed, there has been a few issues with the publishing of exercises-pureconfig. It seems like the scala jar is not on the classpath: https://github.com/scala-exercises/exercises-pureconfig/runs/966737129?check_suite_focus=true#step:9:292.

BenFradet:
ah that might be due to incompatibilities between 2.12 and 2.13 artifacts.

BenFradet:
Maybe we are in the following case: https://github.com/scala/bug/issues/11832#issuecomment-578517653, I noticed that sbt-exercise has 2.13 in `crossScalaVersions`:

https://github.com/scala-exercises/sbt-exercise/blob/master/build.sbt#L32

wdyt @juanpedromoreno ?

juanpedromoreno:
I think the last releases failed because of [this issue](https://github.com/scala-exercises/exercises-cats/issues/134#issuecomment-671493297). I couldn't spend time on it, but I hope to do it shortly.

The problem with the pureconfig exercises would be elsewhere, probably because the scaladoc comments are not completely valid, but couldn't figure it out when I looked into it :( .

paualarco:
Ok I see... thought that the last time the pipeline was not even succeeding. So I guess we should fix the release for `exercises-pureconfig` first.
It's strange that the [Eval](https://github.com/scala-exercises/exercises-cats/pull/87) exercises were successfully released whereas the [OptionT](https://github.com/scala-exercises/exercises-cats/pull/88) that was added one commit later did never get released...

